PR #1, opened 2023-10-03
ITM-60

dereknop:
To test go to http://localhost:3000/test. All other mvp code was left alone.


dereknop:
> To test go to http://localhost:3000/test. All other mvp code was left alone.

No longer test. http://localhost:3000/actionsCompare

dashboard-ui/public/civilianBob.png
brianpippin, line 1: Eventually all these image files should be in a location separate from the dashboard.   Then hopefully we can ingest or add a file location when putting this data into the database and the UI would use that data to look up the correct file, but we need to improve our data ingestion process first.

dashboard-ui/public/jungleExampleData/human1.csv
brianpippin, line 1: One of the next highest priority things we should be looking to do is creating an ingest tool that will take csv/json and put it into the database.  Here's an example of something I created for MCS for inspiration:

https://github.com/NextCenturyCorporation/mcs-ingest

dashboard-ui/public/jungleExampleData/jungle1.html
brianpippin, line 1: Ideally we work with Jacob to move any data that you need that was in these html files into the JSON output he is adding to the simulator so we don't need these html files.

dashboard-ui/src/components/ActionsCompare/actions.jsx
```diff
@@ -0,0 +1,108 @@
...
+class Actions extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            selectedScenario: '', // Initially set to 'disabled'
+        };
+    }
+
+    handleScenarioChange(target) {
+        //const selectedScenario = e.target.value;
```
brianpippin, line 30: We should remove this commented out line.

mongo_backup.tar.gz
brianpippin, line 1: We shouldn't commit this file.  It should be removed from the repo.

dashboard-ui/src/components/Results/resultsTable.jsx
```diff
@@ -0,0 +1,247 @@
...
+                                    if (error) return <div>Error</div>
+                                    const testData = data[getTestByADMandScenarioName];
+
+                                    return (
+                                        <>
+                                            {testData !== null && testData !== undefined &&
+                                            <>
+                                                <AlignmentScoreBox performer={this.formatADMString(this.state.adm)} data={testData} scenario={this.state.scenario}/>
+                                                <div className='paper-container'>
+                                                    <TableContainer style={{ maxHeight: '70vh' }}>
```
brianpippin, line 211: We should move to a css class.  (Know it's not yours but all the files are in the code review so checking them all)

dashboard-ui/src/components/Home/scoreChart.jsx
```diff
@@ -0,0 +1,114 @@
...
+    );
+};
+
+class ScoreChart extends React.Component {
+
+    constructor(props) {
+        super(props);
+
+        this.state = { 
+            style: {fontFamily: "Lato", textAlign: "center"}
```
brianpippin, line 66: So it would be interesting to see if this style is being used, because, I see we are using state.styles so perhaps both can be removed, or perhaps both used, either way, we should check that out.

dashboard-ui/src/components/ActionsCompare/scenarioDetails.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+    componentDidUpdate(prevProps) {
+        if (this.props.selectedScenario !== prevProps.selectedScenario) {
+            this.setState({ scenarioId: this.props.selectedScenario })
+        }
+    }
+
+    render = () => {
+        return (
+            <Accordion defaultActiveKey="0">
```
brianpippin, line 57: It would be nice to remove the Accordion from the top level of the Scenario Details, but keep the Accordion for the patients.   I think having a few details always visible would be good for the page.

dashboard-ui/src/components/ActionsCompare/decisionMakerDetails.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+    render = () => {
+        return (
+                <Row>
+                    <Col>
+                        <Tabs className="pt-1 px-1">
+                            <Tab eventKey="0" title="Action List" className="p-4">
+                                <DecisionList title={this.state.dropdownLabel} decisionMaker={this.state.decisionMaker} isHuman={this.state.isHuman} selectedScenario={this.props.selectedScenario}/>
+                            </Tab>
+                            {/* commenting out the scenario stats until I am parse the adm data to match the same format
```
brianpippin, line 65: Ideally we have a way to get the scenario stats into the database and then use a query to pull them out.

dashboard-ui/src/components/ActionsCompare/decisionMakerDash.jsx
```diff
@@ -0,0 +1,117 @@
...
+        if (this.props.decisionMaker !== prevProps.decisionMaker) {
+            this.setState({dashInfo: null})
+            this.fetchData(this.props.decisionMaker);
+        }
+    }
+
+    fetchData() {
+        if (this.props.decisionMaker && this.props.isHuman) {
+
+            const htmlFilePath = process.env.PUBLIC_URL + `/jungleExampleData/${this.props.decisionMaker === "human1" ? "jungle1" : "jungle2"}.html`;
```
brianpippin, line 23: I know I mentioned this somewhere else, but ideally the data that is in these html files is in the JSON files and we then ingest that into the mongo database and then use an endpoint here to get that data.

dashboard-ui/src/components/ActionsCompare/decisionList.jsx
brianpippin, line 1: This file is pretty large for a react file, would like to see it broken into smaller components if possible.

PR #2, opened 2023-11-28
Itm 135

dereknop:
Relies on the aws-s3 branch of osu-simulator to upload simulator run data to mongodb docker container. The dashboard pulls this data and displays it on the action compare page. Additions were made to graphql server.schema file to query the necessary data. Cleared out any local files that were being used for human data. Refactored old rendering functions that were using csv/html files for humans to render using the updated JSON data from osu-sim (images rendered from bytes).


nextcen-dgemoets (commented):
I can't properly review this because I still can't run Docker on Windows...but I did have a comment about what could be a spurious file/reference.

dashboard-ui/src/components/Survery/survery.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+class Survery extends React.Component {
```
kaitlyn-sharo, line 3: Purely out of curiosity since I might be working on the dashboard one day, is there a reason we're mixing up the two different ways of defining components? I've seen both ways in this code base (class with render and a function (which is my preferred way)) so I just wanted to know why you choose each one when you do.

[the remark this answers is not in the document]
dereknop: That is a typo, thanks.

dashboard-ui/src/components/App/index.jsx
```diff
@@ -4,6 +4,7 @@ import ResultsPage from '../Results/results';
 import HomePage from '../Home/home';
 import ScenarioPage from '../ScenarioPage/scenarioPage';
 import Actions from '../ActionsCompare/actions';
+import Survey from '../Survery/survery';
```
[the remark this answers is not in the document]
dereknop: That component will be used for the delegation survey. I did not mean to include that in this PR at all (which is why it isn't used in the file).

dashboard-ui/src/components/Survery/survery.jsx
dereknop: I started working on the dashboard based off of MVP1 code and all of the components I have made I have defined this way. So the other way to define a component you are referring to was probably written by either Jacob or Brian. I am completely fine with standardizing to whatever way you prefer (I have no preference). This is just the way I learned how to define React components when I started picking up React a couple months ago.

dashboard-ui/src/components/App/index.jsx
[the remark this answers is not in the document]
dereknop: I'll just remove that component and reference from this PR as it is not relevant to the changes I was making

dashboard-ui/src/components/Survery/survery.jsx
kaitlyn-sharo: They can be mixed without a problem, I just didn't know if there was a reason. Since there's not, no big deal. In my opinion, functionally defined components are easier to program, read, and control and render faster. I also think that some functionality that is available in functional components is not available in class components, or is at least much easier, but I'm not trying to turn your world upside down :)

dereknop: No problem at all, thanks for pointing that out. I'm curious to look into it, especially if it your approach renders faster.

kaitlyn-sharo: https://react.dev/reference/react/Component This React official link recommends using functions over classes, but you should probably do some more external research on why because they don't really explain why

brianpippin: I prefer the classes just on a personal note, I feel like it is easier to keep track of the state object, but then again the function way also gets you lots of functionality and I have used both before.  My 2 cents.

dashboard-ui/src/components/ActionsCompare/decisionList.jsx
```diff
@@ -86,196 +67,38 @@ class DecisionList extends React.Component {
...
-  }
-
-
   filterDecisionsSoarTech(decisions) {
     // filter out decisions that involve bbn scenario 
-    const filteredNames = [
-      "Civilian Bob_4 Root",
-      "Military Mike Jungle Combat_1_5 Root",
-      "Military Mike Jungle Scout_1_3 Root"
+    const relevantCasualties = [
```
brianpippin, line 79: At some point we will want this array and the array at line 95 to be dynamic and not hard coded in here.

PR #3, opened 2023-12-21
Itm 143

dereknop:
Delegator survey. Posts to dashboard-mongo on completion. Measures time spent on each page. Randomized order in which the different presentation styles are shown. Randomized order in which the two different decision makers in each presentation style are shown.


brianpippin (approved):
Everything looks good, I didn't get a chance to look at the database data format, if you just want to send that over that would be good, although I am sure if it is json and all the data is there it will be good.

PR #7, opened 2024-01-29
ITM 197 - visualize survey results

kaitlyn-sharo:
To test, clear out your mongo database of survey results, as this uses a new format, and take the survey at least once, then go to the survey results page.

Right now, the visualization is very simplistic. It shows, by scenario, how many people answered each choice for each question. Ewart will hopefully be providing more guidance later this week. I think he may have also wanted to see who answered what choices specifically based on who they delegated to, so that may be a next step. 

**Goal of this PR:** to get a review of the initial survey results format and to encourage new ideas and tasks based on the endless possibilities of analyzing data

### DIFF
--- a/dashboard-ui/src/components/Survey/survey.jsx
+++ b/dashboard-ui/src/components/Survey/survey.jsx
@@ -4,6 +4,8 @@ import { Model } from 'survey-core'
 import { Survey } from "survey-react-ui"
 import surveyConfig from './surveyConfig.json';
 import surveyTheme from './surveyTheme.json';
+import { StaticTemplate } from "./staticTemplate";
+import { DynamicTemplate } from "./dynamicTemplate";
 import gql from "graphql-tag";
 import { Mutation } from '@apollo/react-components';
 import { StaticTemplate } from "./staticTemplate";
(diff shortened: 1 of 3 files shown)


dashboard_ui.sh
brianpippin, line 1: You might need to rebase with main to get rid of these files in here.

brianpippin: Never mind you branched off of Derek's branch.

kaitlyn-sharo: I think this is just a chmod because I couldn't run them

brianpippin (approved):
Code looks good and Page looks good in the UI, some follow on tickets:

1.  We are going need a way to group experiments, so something is going to need to potentially be added to the DB and then a drop down so you can switch experiments.
2.  We want to see what else Ewart wants to add, see if he wants to add the table that Derek was  sending him, that way he could come here and copy out the data.
3.  Meet with Jennifer about what else she might want on this page, but let's get this merged for Friday, and create these follow on tickets.

kaitlyn-sharo:
> Code looks good and Page looks good in the UI, some follow on tickets:
> 
> 1. We are going need a way to group experiments, so something is going to need to potentially be added to the DB and then a drop down so you can switch experiments.
> 2. We want to see what else Ewart wants to add, see if he wants to add the table that Derek was  sending him, that way he could come here and copy out the data.
> 3. Meet with Jennifer about what else she might want on this page, but let's get this merged for Friday, and create these follow on tickets.

Sounds good. I am currently working on #2 by adding a visual table with a download button that will download an excel spreadsheet similar to what Derek has been sending.

PR #9, opened 2024-02-02
Itm 197 Tabulated Survey Results

kaitlyn-sharo:
Added a tabulated view for the survey results with an option to download. Also added filter to table for survey version. In the future, depending on features requested we could easily use a table library. I've often found that table libraries are slow to render/update and difficult to customize which is why I started with vanilla html. Of course, most table libraries would come with additional filtering options and sorting, but I imagine those dealing with the data will just be using the excel file for that.


dereknop:
The table view looks great, and the download worked as expected. I think it does a good job of handling partial survey results (which will hopefully be rare). In the image I'm attaching I wonder if there is a better approach to the unanswered fields than having them blank. Maybe we could just put a "-" in the cell or "No Response". I think it looks good as is, but it might make it more clear for Ewart that the user didn't make it through the survey as opposed to us missing data. 
Screenshot 2024-02-02 at 10 11 09 AM

kaitlyn-sharo:
>  I wonder if there is a better approach to the unanswered fields than having them blank. Maybe we could just put a "-" in the cell or "No Response".

That's a great idea. I will do that as soon as I finish with the validator.

dereknop (approved):

brianpippin (approved):

PR #10, opened 2024-02-07
Dynamic template refinement

dereknop:
- Made vitals into a dropdown to help with the sizing of patient cards
- ^dropdown is a new component
- Cleaned up dynamicTemplate. Moved css styles out of the component, into tempalte.css. 
- Randomization order of medic comparisons and randomization of whether they are dynamic or static should be fully working. 
- Improved layout of dynamic template patient cards


kaitlyn-sharo (approved):
Looks good, works well. I don't like that it adds so much extra clicking, but that's just what comes as a side effect of drop downs. I'm guessing the people who asked for this wanted the users specifically to click the drop downs, and we can't have them automatically expanded (on bigger screens, leave them closed on small screens that might have size problems, but if they're closed, what's the point of having drop downs because eventually they'll open and then the cards will be big again...)?

brianpippin (approved):

PR #14, opened 2024-02-20
Itm 223

dereknop:
The big change here is the addition of the omnibus presentation style. After the existing presentation style of medics, you should now see a medic with three of their scenarios all displayed together. Things to test:

- The three Dynamic templates embedded in the omnibus still work as they did without being grouped as part of three (all of the buttons still work to expand/collapse)
- There are four sets of scenarios but we randomly select three sets to be shown before the omnibus
- Each of the scenarios mentioned have two different medics i.e Medic 33 vs Medic 44. Which pairing of three that make up the omnibus is random (although they obviously can't have two different medics that share a scenario)
- There should be a new question on all of the medic comparison pages that forces you to choose a medic you'd delegate to if you had no choice but to pick one.
- Multiple scenarios in the omnibus can be expanded at the same time

### DIFF
--- a/dashboard-ui/src/components/Survey/dynamic.jsx
+++ b/dashboard-ui/src/components/Survey/dynamic.jsx
@@ -18,14 +18,12 @@ const Dynamic = ({ patients, situation, supplies, decision, dmName, actions, exp
 
     // log actions
     const logAction = (actionName) => {
-        const newLog = { actionName, timestamp: new Date().toISOString() };
-        setActionLogs(prevLogs => {
-            const updatedLogs = [...prevLogs, newLog];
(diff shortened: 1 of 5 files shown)


brianpippin:
Nothing with the code stuck out to me, will build and look it over as well.

brianpippin (approved):
Took the survey, results showed, had no issues.

brianpippin:
Logging fix code looks good.

nextcen-dgemoets (approved):
Well, I still can't quite run the dashboard even with Docker Desktop finally installed, and I don't know React, but I didn't see anything obvious when looking through the code.

brianpippin (approved):

PR #19, opened 2024-02-26
Don't pull survey results with "test" in participantID

dereknop:
This was originally a branch off of ITM-197-Tabulated but it looks like the order in which things got merged was ITM-197-Tabulated into main and then this change into ITM-197-Tabulated. So this change never made it to main. It filters out the surveyResults collection to exclude any entries that have a participantId that contains the word "test"


dereknop:
You may need to rebuild graphql container to test if you have one running already

brianpippin (approved):

PR #21, opened 2024-02-29
Itm 243

dereknop:
Safeguards for duplicates being added to omnibus. Making sure there's no way for anything other than 3 groupedDMs. Log user browser info, OS, platform etc.


brianpippin:
Code looks good, still testing.  Jennifer and I are thinking the Omnibus might look better to the user if we left the number off the scenario in the display, so instead of "Scenario 4", "Scenario 1", "Scenario 3", each one just said "Scenario" on the title bar in the UI.

brianpippin (approved):

PR #24, opened 2024-03-06
Itm 218

dereknop:
There is no link to the page on the navbar because that change is already made in the pr "updated navbar". 
http://localhost:3000/text-based
Uploads to collection "userScenarioResults" (attached to answer is the scenario name i.e "ADEPT Desert Scenario "


brianpippin (commented):
No issues yet.  I did move the "Exit Scenario" button, should I find no more issues I am going to push later tonight.

PR #27, opened 2024-03-07
ITM-15 added nginx config for Eval API

phile-caci:
tested in prod, config is there now if you want to test it.  but an eval dashboard release before this get's in will overwrite the config.

TA3 Eval API Access Point:
https://darpaitm.caci.com/ui

Tested API to make sure it worked.

Ran(through the swagger endpoint): https://darpaitm.caci.com/ta2/startSession?adm_name=phi&session_type=eval&kdma_training=false

Got response: "b09b67d2-c428-4381-810e-71c70c63ab07"

### DIFF
--- a/docker_setup/resources/nginx.conf
+++ b/docker_setup/resources/nginx.conf
@@ -56,7 +56,28 @@ http {
       proxy_set_header 	Upgrade 	  $http_upgrade;
     }
 
+    location ^~ /ui/openapi.json {
+      set $upstream http://itm-server:8080/openapi.json;
+      proxy_pass $upstream;
+    }
+
+    location ^~ /openapi.json {
+      set $upstream http://itm-server:8080/openapi.json;
+      proxy_pass $upstream;
+    }
+
     location /ui/ {
+      set $upstream http://itm-server:8080/$request_uri;
+      proxy_pass $upstream;
+
+      proxy_set_header  Host              $host/ui;   # required for docker client's sake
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (commented):
The chances look good to me.  As we discussed in Teams, I tried running one of the client ADMs against this endpoint, but I couldn't seem to reach it.

Also, obviously this is the `main` branch of the server (so pre-metrics eval), but I have no reason to think this configuration wouldn't work with what we'll be pushing soon for the metrics evaluation.

brianpippin:
I think that is sufficient but we will want to retest once we push the Metrics Evaluation version of the server to main.  Perhaps we can create a ticket that runs adds a test to the evaluation server code that reaches out and hits it so we know anytime it stops working or maybe that test should actually be in the client code.  I just think an automated test would help ensure we don't break anything.

nextcen-dgemoets:
> Perhaps we can create a ticket that runs adds a test to the evaluation server code that reaches out and hits it so we know anytime it stops working or maybe that test should actually be in the client code. I just think an automated test would help ensure we don't break anything.

I could add a `ping` endpoint to the server that just returns success if it's ready to accept new sessions.
Currently, you can just send a `new_session` request as a form of ping.  I guess we can have a cron job or something that requests a new session every few hours, and does something if it gets an error.  But keep in mind there's a hard cap on the number of sessions the server will maintain at a time before they expire-- a cap which I intend to increase because I think the footprint of a session is pretty small.

brianpippin (approved):

PR #28, opened 2024-03-12
Itm 259

dereknop:
Text-based scenarios. 
Enter participant id
Multi-select for what scenarios were completed in VR (if any)
The moderator chooses the order of text-based scenarios for them 
- Scenarios append continuously
- Write as four separate files to MongoDB


dereknop:
> Text-based scenarios. Enter participant id Multi-select for what scenarios were completed in VR (if any) The moderator chooses the order of text-based scenarios for them
> 
> * Scenarios append continuously
> * Write as four separate files to MongoDB

May need to rebuild graphql container for change to mutation to be in effect

PR #33, opened 2024-03-15
Dm comparison modal

dereknop:
On the pages where you are asked to compare two decision makers you just saw you can now refer back to that DM's actions

### DIFF
--- a/dashboard-ui/src/components/Survey/comparison.jsx
+++ b/dashboard-ui/src/components/Survey/comparison.jsx
@@ -1,0 +1,129 @@
+import React from "react";
+import { ElementFactory, Question, Serializer } from "survey-core";
+import { SurveyQuestionElementBase } from "survey-react-ui";
+import { Button, Modal } from "react-bootstrap";
+import Dynamic from "./dynamic";
+import surveyConfig from './surveyConfig.json';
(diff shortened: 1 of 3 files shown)


nextcen-dgemoets:
I think I got to the page and saw the comparison view.  However when I clicked the Back button (as if I thought that would get me back to where I was before I brought up the comparison view), it took me back to an empty form where I'm supposed to enter a Participant ID.  If that's expected, then maybe we need some kind of instruction (or maybe it was there and I didn't see it) not to use the Back button (which remains the bane of any web developer's existence).

dereknop:
> I think I got to the page and saw the comparison view. However when I clicked the Back button (as if I thought that would get me back to where I was before I brought up the comparison view), it took me back to an empty form where I'm supposed to enter a Participant ID. If that's expected, then maybe we need some kind of instruction (or maybe it was there and I didn't see it) not to use the Back button (which remains the bane of any web developer's existence).

Adding ticket to address a user hitting back on the browser

nextcen-dgemoets (approved):
The new pop-up looked good.  Thank you for writing [ITM-285](https://nextcentury.atlassian.net/browse/ITM-285).

PR #34, opened 2024-03-18
Itm 285

dereknop:
Users shouldn't be able to leave using the back arrow on the browser on survey or text-based scenarios without being prompted for confirmation that they want to leave. The text-based scenario doesn't confirm until they are actually in one of the scenarios (no restriction from leaving the landing page for text-based scenarios).

### DIFF
--- a/dashboard-ui/src/components/Survey/survey.jsx
+++ b/dashboard-ui/src/components/Survey/survey.jsx
@@ -12,6 +12,7 @@ import gql from "graphql-tag";
 import { Mutation } from '@apollo/react-components';
 import { getUID, shuffle } from './util';
 import Bowser from "bowser";
+import { Prompt } from 'react-router-dom'
 
 const UPLOAD_SURVEY_RESULTS = gql`
   mutation UploadSurveyResults( $surveyId: String, $results: JSON) {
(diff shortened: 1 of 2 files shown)


jaudick (approved):
I cant run it but with my react knowledge this looks good

nextcen-dgemoets (changes requested):
Looks good.  My only suggestion would be to tweak the wording to clarify that they'll lose all progress.  Maybe, "Please finish the survey before leaving the page. By hitting 'OK' you will be leaving the survey before completing, losing all progress."   (or "and will have to start over" or something like that).

nextcen-dgemoets:
Interested in @brianpippin thoughts on a possible wording change, though.

dereknop:
> Looks good. My only suggestion would be to tweak the wording to clarify that they'll lose all progress. Maybe, "Please finish the survey before leaving the page. By hitting 'OK' you will be leaving the survey before completing, losing all progress." (or "and will have to start over" or something like that).

Only reason I didn't say they would lose all progress is because we it uploads progress to mongo as they go not just at the end (was requested by OSU a while ago). But lying to them to scare them may not be a bad idea haha

brianpippin:
Looks good, I don't want to say lose all progress.  I think this would be sufficient:

"Please finish the survey before leaving the page. By hitting 'OK' you will be leaving the survey before completion and will be required to start the survey over at the beginning."

nextcen-dgemoets:
> "Please finish the survey before leaving the page. By hitting 'OK' you will be leaving the survey before completion and will be required to start the survey over at the beginning."

Looks good to me.

@dereknop Presumably if they start over it will overwrite whatever has already been written to the DB (assuming they use the same participant ID)?

PR #36, opened 2024-03-20
Itm 249 text results

kaitlyn-sharo:
Added text results page:
- Chart view with percentages of all answers (including answers no participant chose)
- Text view (really tables) with count/percentages of all answers (default, because Jennifer likes this view the most)
- Participant view - one big table for each scenario containing a participant-indexed view of responses


brianpippin:
Looks great!  The page is amazing.  I just have some minor things.  First I think this should be its own component folder like the rest of the pages.  So I would create a new folder in 'components' maybe 'TextBasedResults' and move 'TextBasedResultsPage.jsx' under that.   I would then create a new css file and add that in that folder as well and move the css you added under that.  That way all of that stays separate.

Also I am going to be adding some of this to Derek's code reviews as well, but I would like to see to start to work on the console warnings some as well (I think one of them might be old and be mine), but here are the warnings for the text results page:

webpackHotDevClient.js:115 ./src/components/SurveyResults/TextBasedResultsPage.jsx
  Line 147:8:   React Hook React.useEffect has a missing dependency: 'pageName'. Either include it or remove the dependency array      react-hooks/exhaustive-deps
  Line 174:8:   React Hook React.useEffect has a missing dependency: 'pageName'. Either include it or remove the dependency array      react-hooks/exhaustive-deps
  Line 242:8:   React Hook React.useEffect has a missing dependency: 'scenarioName'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
  Line 282:22:  'error' is assigned a value but never used                                                                             no-unused-vars
  Line 427:85:  Expected to return a value at the end of arrow function      
  
 Everything looks great though, love the new page.

kaitlyn-sharo:
@brianpippin Thanks for the feedback! Everything should be fixed now

PR #38, opened 2024-03-29
Itm 304

dereknop:
Add alignment scores to all preexisting data. 

Inside of TextBasedScenarios run the following to test:
`node mapData.js <mongo username> <mongo password>`

note that mongo is now running on port 27030 to avoid conflict with soartech TA1 server mongo container


node-graphql/server.mongo.js
```diff
@@ -20,7 +20,7 @@ let AWS = require('aws-sdk');
 // });
 
 // We connect mongoose to our local mongodb database
-const connection = mongoose.connect('mongodb://dashboard-mongo:27017/dashboard', {
+const connection = mongoose.connect('mongodb://dashboard-mongo:27030/dashboard', {
```
brianpippin, line 23: If you are changing the port in this file, you probably also need to modify the prod docker-compose port as well.

dashboard-ui/src/components/TextBasedScenarios/TextBasedScenariosPage.jsx
```diff
@@ -59,19 +60,18 @@ class TextBasedScenariosPage extends Component {
...
         const scenarioOrderArray = JSON.parse(scenarioOrderString);
-        console.log(scenarioOrderArray)
         this.setState({
             scenarios: scenarioOrderArray,
             participantID: survey.data["Participant ID"],
             vrEnvCompleted: survey.data["vrEnvironmentsCompleted"]
         })
 
-        const selectedScenarios = []
+        let selectedScenarios = []
```
kaitlyn-sharo, line 71: Why was this changed to let? I don't see a reassignment of it

dashboard-ui/src/components/TextBasedScenarios/TextBasedScenariosPage.jsx
```diff
@@ -59,19 +60,18 @@ class TextBasedScenariosPage extends Component {
...
         for (const scenario of scenarioOrderArray) {
-            selectedScenarios.push(scenarioMappings[scenario])
+            // make deep copies of json files to be sure the originals are not unintentionally tampered with
+            selectedScenarios.push(JSON.parse(JSON.stringify(scenarioMappings[scenario])))
```
kaitlyn-sharo, line 74: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

I wonder if you'd be able to use this here?  It's pretty new, I just learned about it last year, but it cleans up the whole JSON parse/stringify method of cloning

dashboard-ui/src/components/TextBasedScenarios/TextBasedScenariosPage.jsx
```diff
@@ -201,6 +200,32 @@ class TextBasedScenariosPage extends Component {
...
+                        console.err("Error mapping user selection to choice ID");
+                    }
+                } else {
+                    choice = "choice-"+indexOfAnswer;
+                }
+                question[1].choice = choice;
+            })
+        })
+    }*/
+
```
kaitlyn-sharo, line 228: Is this intentionally left here?

kaitlyn-sharo (approved):
Can't test with soartech until I get reinvited to their repo, but adept works great and code looks good (except for the few little things in my comments)

dashboard-ui/src/components/TextBasedScenarios/TextBasedScenariosPage.jsx
dereknop: I think just leftover from when I was debugging, I'll change it back

dashboard-ui/src/components/TextBasedScenarios/TextBasedScenariosPage.jsx
dereknop: Well ideally I'd like to mark the choice id to the documents before they get uploaded to mongo instead of having to write the script so that is the purpose of this being here. This will get expanded on so I think I'll just leave it for now

kaitlyn-sharo: Great, I just wanted to make sure :)

dashboard-ui/src/components/TextBasedScenarios/util.js
```diff
@@ -0,0 +1,183 @@
...
+                question[1].probe = pageQuestion.probe;
+            }
+        })
+    })
+}
+
+const getAdeptAlignment = async (scenarioResults, scenarioId) => {
+    /*local host 8080 is where I had adept running, if you had to change your port must change script to match */
+    // build fetch for starting new session
+    const url = 'http://127.0.0.1:8080/api/v1/new_session';
```
brianpippin, line 51: So this script isn't going to work on production because the TA1 servers are on different machines.  I think we need to get prod urls to test this, and we might need to be on the VPN to do it.

dashboard-ui/src/components/TextBasedScenarios/TextBasedScenariosPage.jsx
dereknop: Thanks, I'll bookmark this page and be sure to test that and use it in my next PR. Just don't want to accidentally break something right now

PR #39, opened 2024-03-29
Downloadable Aggregate Results

kaitlyn-sharo:
To test, run Derek's scripts to update the database with alignment scores. Also, run the probe_matcher.py script from the itm-ingest repo to get simulation alignment scores to the database. View the new Aggregated Results page and press the download button. More tables/graphs on the page will come once Jennifer has created the final list


dashboard-ui/src/components/App/index.jsx
```diff
@@ -149,9 +150,14 @@ export class App extends React.Component {
...
                                 </li>
-                                <li className="nav-item">
-                                    <Link className="nav-link" to="/results">Results</Link>
-                                </li>
+                                <NavDropdown title="Results">
+                                    <NavDropdown.Item as={Link} lassName="dropdown-item" to="/results">
+                                        Alignment Results
+                                    </NavDropdown.Item>
+                                    <NavDropdown.Item as={Link} lassName="dropdown-item" to="/dataset">
+                                        Aggregate Results
```
brianpippin, line 158: Can you change results back to how it was and add something on the end of the nav that says "Data Analysis" or something like that.  I want to keep your aggregate results separated because it will include all the other drop down results in yours.  Delegation/ADM/Text based

dereknop (approved):
Works great, got it showing alignment data from all the text-based results. The only thing I'd say is that it might not be obvious to someone examining data what the significance of the integer value for "PostVRState" is.

kaitlyn-sharo:
> Works great, got it showing alignment data from all the text-based results. The only thing I'd say is that it might not be obvious to someone examining data what the significance of the integer value for "PostVRState" is.

True, there's a lot of those, but Jennifer gave me the specific calculations and maps for everything, so she knows exactly what she wants. It's all in that document I sent you earlier this week, if you're curious

dashboard-ui/src/components/App/index.jsx
kaitlyn-sharo: Should be fixed now

PR #40, opened 2024-03-29
ITM-310: Modify the dashboard to be able to display things per evaluation

brianpippin:
You need to have run all the ingest scripts for this to work, as this needs the script that adds evaluation Name and numbers to every record.  This will allow the home page to view the charts by evaluation, it will allow the adm results and adm scenario pages to let you pick which evaluation you would like to view first.

### DIFF
--- a/dashboard-ui/src/components/App/index.jsx
+++ b/dashboard-ui/src/components/App/index.jsx
@@ -146,16 +146,16 @@ export class App extends React.Component {
                                     <Link className="nav-link" to="/">Home</Link>
                                 </li>
                                 <NavDropdown title="ADM Results">
-                                    <NavDropdown.Item>
-                                        <Link className="dropdown-item" to="/results">Results by ADM</Link>
+                                    <NavDropdown.Item as={Link} className="dropdown-item" to="/results">
+                                        Results by ADM
                                     </NavDropdown.Item>
-                                    <NavDropdown.Item>
-                                        <Link className="dropdown-item" to="/scenarios">ADM Scenarios</Link>
+                                    <NavDropdown.Item as={Link} className="dropdown-item" to="/scenarios">
+                                        ADM Scenarios
                                     </NavDropdown.Item>
                                 </NavDropdown>
                                 <NavDropdown title="Delegation Surveys">
-                                    <NavDropdown.Item>
-                                        <Link className="dropdown-item" to="/survey">Take Survey</Link>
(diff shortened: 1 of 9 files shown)


kaitlyn-sharo (approved):
The organization of these pages makes sense to me! 

I had to make some changes to get everything on the right port after Derek's changes, but all of that will be fixed when everything is merged.

I still wonder if on some of these pages we can some day auto-select the latest eval and first of each subsequent category from it so there's less clicking to get to a non-empty page, but that's a later issue.

dashboard-ui/src/components/App/index.jsx
```diff
@@ -150,16 +144,15 @@ export class App extends React.Component {
...
-                                <NavDropdown title="Surveys">
+                                <NavDropdown title="ADM Results">
+                                    <NavDropdown.Item>
+                                        <Link className="dropdown-item" to="/results">Results by ADM</Link>
+                                    </NavDropdown.Item>
+                                    <NavDropdown.Item>
+                                        <Link className="dropdown-item" to="/scenarios">ADM Scenarios</Link>
+                                    </NavDropdown.Item>
+                                </NavDropdown>
+                                <NavDropdown title="Delegation Surveys">
```
kaitlyn-sharo, line 155: Forgot to mention this in my review. This causes errors in the console every time you click on a nav. An easy way to remedy this is add as={Link}, the `to` field and the `className` field to NavDropdown.Item, then just remove link completely. Examples should be below this piece of code where I've made this fix

dereknop (approved):
I did not run into the issue Kaitlyn had with the empty tabs 
Screenshot 2024-03-29 at 4 27 12 PM

I have attached a screen recording of the ID and name in the response column getting chopped off when you scroll to side. Since there is so much space there, I don't think this is the expected behavior.

https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/assets/73036954/da917f88-cd8a-424b-84e6-c843dfc08c39

brianpippin:
> I did not run into the issue Kaitlyn had with the empty tabs Screenshot 2024-03-29 at 4 27 12 PM
> 
> I have attached a screen recording of the ID and name in the response column getting chopped off when you scroll to side. Since there is so much space there, I don't think this is the expected behavior.
> 
>  Screen.Recording.2024-03-29.at.4.31.15.PM.mov

Derek - I didn't change of that area, that is all carried over from MVP and the September Milestone but I think we need a ticket to go back and clean up that area.

PR #45, opened 2024-04-08
Itm 336

dereknop:
Addressed console errors from surveyResults, resultsTable, and util.js script. 
Included URLs for both prod. and dev. for alignment scores script.

### DIFF
--- a/dashboard-ui/src/components/TextBasedScenarios/util.js
+++ b/dashboard-ui/src/components/TextBasedScenarios/util.js
@@ -1,5 +1,10 @@
 const axios = require('axios');
-const { stUrban, stDesert, stJungle, stSubmarine } = require('./problemProbes')
+const { stUrban, stDesert, stJungle, stSubmarine } = require('./problemProbes');
+// for env variables
+require('dotenv').config();
+const ADEPT_URL = process.env.ADEPT_URL;
+const ST_URL = process.env.ST_URL;
+
 // add proper choice id's to questions
 const mapAnswers = (scenario, scenarioMappings) => {
     if (!scenario.participantID) { return; }
@@ -46,15 +51,8 @@ const addProbeIDs = (scenario, scenarioMappings) => {
 }
 
 const getAdeptAlignment = async (scenarioResults, scenarioId) => {
-    /*local host 8080 is where I had adept running, if you had to change your port must change script to match */
     // build fetch for starting new session
-    
-    //PROD
-    const url = 'http://10.216.38.70/api/v1/new_session';
-
-    // DEV
-    // const url = 'http://127.0.0.1:8080/api/v1/new_session';
-    
+    const url = ADEPT_URL + '/api/v1/new_session';
     try {
         const startSession = await axios.post(url);
         const sessionId = startSession.data;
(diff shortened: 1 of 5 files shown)


dashboard-ui/src/components/TextBasedScenarios/util.js
```diff
@@ -94,30 +107,37 @@ const getAdeptAlignment = async (scenarioResults, scenarioId) => {
 const getSoarTechAlignments = async (scenarioResults, scenarioId) => {
     /*local host 8084 is where I had soartech running, if you had to change your port must change script to match */
     // build fetch for starting new session
+    // PROD
     const url = 'http://10.216.38.125:8084/api/v1/new_session?user_id=default_use';
 
+    // DEV
+    // const url = 'http://127.0.0.1:8084/api/v1/new_session?user_id=default_use';
```
kaitlyn-sharo, line 114: I think all of these //Prod //Dev sections would be better handled by two global variables at the top of the util file, if not in a .env file. Then you just need the first half "http://127.0.0.1:8080/4" since the rest of the url is the same. For example:

const SOARTECH_URL = "http://127.0.0.1:8080"

line 113 -> const url = SOARTECH_URL+"/api/v1/new_session..."

That way there's not as much commented code to toggle back and forth between. You could even optionally have DEV_SOARTECH_URL and PROD_SOARTECH_URL (i'm just using soartech as an example, this of course applies to adept as well), and then just set a .env REACT_APP_DEV_MODE=true/false which would toggle back and forth between them.

line 113 -> const url = (REACT_APP_DEV_MODE==='true' ? DEV_SOARTECH_URL : PROD_SOARTECH_URL) + 'restoflink'

kaitlyn-sharo (changes requested):
Everything looks good except the utils page with all the commented out links, which can be consolidated into ~2 lines of code. Will test now

dashboard-ui/src/components/TextBasedScenarios/util.js
```diff
@@ -48,17 +48,26 @@ const addProbeIDs = (scenario, scenarioMappings) => {
 const getAdeptAlignment = async (scenarioResults, scenarioId) => {
     /*local host 8080 is where I had adept running, if you had to change your port must change script to match */
     // build fetch for starting new session
+    
+    //PROD
     const url = 'http://10.216.38.70/api/v1/new_session';
```
kaitlyn-sharo, line 53: Runs once I change this line to have :8080

dashboard-ui/src/components/TextBasedScenarios/util.js
dereknop: Good call, added .env file. The user can just toggle back and forth between prod and dev URL from the env file. I did have to add the dotenv package to be able to read from the .env file

kaitlyn-sharo: You don't have to add dotenv if you use React's environment variables. In React, all environment variables accessed through process.env must be prefixed with REACT_APP_. The .env can go at the root level of dashboard-ui and then you shouldn't need any other library or require statement

kaitlyn-sharo (approved):

dashboard-ui/src/components/TextBasedScenarios/util.js
kaitlyn-sharo: ^ignore the above because this is a standalone script (and maybe should be moved elsewhere because of that in the future), not part of the react app, so react variables are not going to work here!

PR #47, opened 2024-04-09
Play by play - Human Sim Data

kaitlyn-sharo:
To test: rebuild to get new query. Check out play-by-play page under human evaluation segments


brianpippin:
Looks great, my only comment is the same one I left on Derek's.  Would love to see us using the scroll on the page, and removing the over flow from table-container although it looks like some other css might need to change to make it happen.  Looks great.

brianpippin:
> Looks great, my only comment is the same one I left on Derek's. Would love to see us using the scroll on the page, and removing the over flow from table-container although it looks like some other css might need to change to make it happen. Looks great.

After some thought - I think it is good to leave as is.

brianpippin (approved):

dereknop (approved):
Looks great

PR #48, opened 2024-04-10
Itm 314

dereknop:
http://localhost:3000/results

adm results table


brianpippin:
I like this is so much better than it was.  I would like to see it just use the scroll on the page instead of giving the internal div overflow, that would be my only suggestion

.css-rorn0c-MuiTableContainer-root {
    width: 100%;
    /* overflow-x: auto; */
}

What what I did to get the desired effect in the browser inspect.

kaitlyn-sharo (approved):
Way better than before

brianpippin (approved):

PR #50, opened 2024-04-11
ITM-178: Add evaluator and admin roles and admin functionality

brianpippin:
NOTE:  You will want to run the script in 'itm-ingest' with your username so you can test the Admin functionality.

1. You should under your profile on the right, now have a drop down item for "Admin"
2. Admins can assign users Admin or Evaluator roles.
3. Only Admins or Evaluators are able to see the data, other users only get the Data Collection items and a very bare home page.
4. You might need to create a local test user to verify 3 on this list

Additional note - going to create a ticket to lock down all the direct URLS with some of this functionality as a follow on ticket.

### DIFF
--- a/dashboard-ui/src/components/Account/adminPage.jsx
+++ b/dashboard-ui/src/components/Account/adminPage.jsx
@@ -15,9 +15,14 @@ const GET_USERS = gql`
 const UPDATE_ADMIN_USER = gql`
     mutation updateAdminUser($username: String!, $isAdmin: Boolean!){
         updateAdminUser(username: $username, isAdmin: $isAdmin) 
-  }`;
+    }`;
 
-function DualInputBox({options, selectedOptions}) {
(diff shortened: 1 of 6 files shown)


dereknop (approved):
Tested as an evaluator and as an admin. Looks good

kaitlyn-sharo (approved):
image

image


Everything seems to work as expected, though I was surprised we have the ability to remove ourselves from the admin list. However, I don't think anybody would do that on accident so it's not a problem.

When I refresh the page, I keep getting the error shown in the screenshot. It doesn't really affect anything; if I press the 'x' everything works as expected, but maybe something to look into? It's possible it's just bad data on my end. I don't know how I'd keep getting that though, as I've wiped my database and started over from mongo's data fairly recently. It looks like it might just need some optional chaining on currentUser and it would probably be fine.

PR #56, opened 2024-04-16
Hotfixes for latest release

phile-caci:
Took me a while to figure out the issue with the release, but in the end the updated react was routing differently from the previous version and I just had to update the NGINX conf.

Then there were some issues with the data on prod, looked like some test data didn't have the alignment/kdma_value keys in the json.  Updated code to handle the "undefined". [REVERTED CHANGE, Kaitlyn already had a fix in dev]

Also on the admin page was broken if user didn't input an email address... checked for "undefined" for the email.  There was a test? user on prod that didn't have an email that was causing an issue there.


dashboard-ui/src/components/AggregateResults/DataFunctions.js
```diff
@@ -429,8 +429,8 @@ function populateHumanDataRow(rowObject) {
         "ST_6.2": "",
         "ST_8.1": "",
         "ST_8.2": "",
-        "AD_KDMA_Env": rowObject[adept].data.alignment.kdma_values.length > 0 ? Math.round(rowObject[adept].data.alignment.kdma_values[0].value * 100) / 100 : "",
-        "ST_KDMA_Env": rowObject[soartech].data.alignment !== null && rowObject[soartech].data.alignment.kdma_values.length > 0 ? Math.round(rowObject[soartech].data.alignment.kdma_values[0].value * 100) / 100 : "",
+        "AD_KDMA_Env": rowObject[adept].data.alignment.kdma_values !== undefined && rowObject[adept].data.alignment.kdma_values.length > 0 ? Math.round(rowObject[adept].data.alignment.kdma_values[0].value * 100) / 100 : "",
+        "ST_KDMA_Env": rowObject[adept].data.alignment.kdma_values !== undefined ? rowObject[soartech].data.alignment !== null && rowObject[soartech].data.alignment.kdma_values.length > 0 ? Math.round(rowObject[soartech].data.alignment.kdma_values[0].value * 100) / 100 : "" : "",
```
kaitlyn-sharo, line 433: There's already a fix for this in development using optional chaining to catch if anything before kdma_values is undefined or null

kaitlyn-sharo (approved):
Looks good

dashboard-ui/src/components/AggregateResults/DataFunctions.js
phile-caci: ah thanks, I'll revert this then

PR #59, opened 2024-04-16
added definitions and changed header/title for aggregate results

kaitlyn-sharo:


dashboard-ui/src/components/AggregateResults/Metrics Eval Variables.xlsx
brianpippin, line 1: I think we want to remove the spaces from the file names, I don't think Windows machines will like that.

dashboard-ui/src/components/AggregateResults/Metrics Evaluation Human Dataset Variables Dashboard.docx
brianpippin, line 1: Same comment as above.

dashboard-ui/src/components/AggregateResults/Metrics Eval Variables.xlsx
kaitlyn-sharo: Updated

brianpippin (approved):
This looks good, Jennifer wanted to double check a couple of columns to make sure they look good, you should drop the export in the chat for her.

dereknop (approved):

PR #60, opened 2024-04-17
Changed AlignScore calculations

kaitlyn-sharo:
Use with branch itm-346 by derek in ingest to get high/low text alignment values

https://github.com/NextCenturyCorporation/itm-ingest/pull/12


brianpippin (approved):

kaitlyn-sharo:
> I crash when trying to go to the text based results participant tab.
> 
> `TextBasedResultsPage.jsx:220 Uncaught TypeError: Cannot read properties of undefined (reading '0')`
> 
> Also, I made a change to itm-346 on ingest to stop a different crash I was experiencing on the page but just to be safe I think we should add optional chaining to line 331 of the same file in case a null value is ever in a mongo document
> 
> `if (typeof (result[k]) === 'object' && result[k]?.questions) {`

Yes, not due to the new calculations in the dashboard, but since alignmentData keeps showing up as null in the database and since we changed the name of alignmentData to high and low. It should work now, with these fixes

dereknop (approved):
Awesome, looks good.

PR #63, opened 2024-04-24
Itm 349

dereknop:
survey config changes for future survey results to be stored corresponding scenario title on each page. Fixed bug in omnibus survey config. Survey Results page for version 2 should now be broken down by scenario title

### DIFF
--- a/dashboard-ui/src/components/SurveyResults/surveyResults.jsx
+++ b/dashboard-ui/src/components/SurveyResults/surveyResults.jsx
@@ -150,7 +150,6 @@ export function SurveyResults() {
         fetchPolicy: 'network-only',
     });
     const [scenarioIndices, setScenarioIndices] = React.useState(null);
-    const [scenarioNames, setScenarioNames] = React.useState(null);
     const [selectedScenario, setSelectedScenario] = React.useState(-1);
     const [resultData, setResultData] = React.useState(null);
     const [showTable, setShowTable] = React.useState(false);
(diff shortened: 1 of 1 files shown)


dereknop:
Note: only survey version two results have scenario titles, version one is still indexed Scenario 1-5 on account of me not knowing if those scenarios even match up to an old TA1 scenario.

dereknop:
> survey config changes for future survey results to be stored corresponding scenario title on each page. Fixed bug in omnibus survey config. Survey Results page for version 2 should now be broken down by scenario title

https://github.com/NextCenturyCorporation/itm-ingest/pull/13

nextcen-dgemoets (approved):
All of the new stuff was there and everything worked and looked good.

dashboard-ui/src/components/SurveyResults/surveyResults.jsx
```diff
@@ -213,7 +215,7 @@ export function SurveyResults() {
                 }
                 for (const x of Object.keys(obj)) {
                     const res = obj[x];
-                    if (res?.scenarioIndex === selectedScenario) {
+                    if (res?.scenarioIndex == selectedScenario) {
```
kaitlyn-sharo, line 218: Was the triple equal removed on purpose?

dashboard-ui/src/components/SurveyResults/surveyResults.jsx
```diff
@@ -76,7 +76,7 @@ function ScenarioGroup({ scenario, data, version }) {
     }, [data]);
 
     return (<div className='scenario-group'>
-        <h2 className='scenario-header'>Scenario {scenario}</h2>
+        <h2 className='scenario-header'>Scenario {scenarioIndices[scenario]}</h2>
```
kaitlyn-sharo, line 79: This duplicates 'scenario' when there is no name since you set the string to "scenario [index]" later on, so it becomes "scenario scenario index".

dashboard-ui/src/components/SurveyResults/surveyResults.jsx
```diff
@@ -184,20 +184,22 @@ export function SurveyResults() {
...
                 }
             }
-            indices = Array.from(new Set(indices));
-            indices.sort((a, b) => a - b);
-            setScenarioIndices(indices);
-            if (indices.length > 0) {
+            const sortedIndices = Object.keys(scenarios).map(Number).sort((a, b) => a - b);
+            setScenarioIndices(scenarios);
+
+            if (sortedIndices.length > 0) {
```
kaitlyn-sharo, line 202: I'm not certain I understand the point of this section. Before, it looks like this was making sure the scenarios would appear in order. Now, it looks like you're sorting the indices just to see if there's at least one key. If you are just looking to see if there's at least one entry in your scenarios object, you should probably just check if Object.keys(scenarios).length > 0. The sorting doesn't seem to be doing anything here, but I might be missing something

kaitlyn-sharo (changes requested):
It does seem to work

dashboard-ui/src/components/SurveyResults/surveyResults.jsx
dereknop: Yes, for the type conversion. Although I could cast this earlier and keep the triple equal

dashboard-ui/src/components/SurveyResults/surveyResults.jsx
dereknop: Thanks for the catch, this should have been changed (leftover from when I was testing). I will change this as the sort doesn't accomplish anything.

dashboard-ui/src/components/SurveyResults/surveyResults.jsx
kaitlyn-sharo: Yeah, probably better to keep types the same

kaitlyn-sharo (approved):

brianpippin (approved):

PR #64, opened 2024-04-25
Bug Fix on Create Account Page

dereknop:
Even though the user was being met with an error message when trying to create an account with no email, a user was still being added to the database with no email. Updated form to validate all fields are provided.


nextcen-dgemoets (approved):
Looks good.  The toast message worked and no item was added to the users database.

One small suggestion-- I'd consider trimming leading/trailing whitespace from the email address.  You'll get a GraphQL error ("Invalid email") so nothing will get created, but if we trim, they won't have to clear the error, realize they had whitespace, and fix it.

brianpippin (approved):

PR #68, opened 2024-05-14
Itm 395

dereknop:
https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-395

https://github.com/NextCenturyCorporation/itm-ingest/pull/17

Started off as bug fix for data not showing up that should be displayed. Fixes 'What is your role' column in survey results table as well as the vr scenarios already completed. Run the new ingest script in the pr I linked to first to test.

### DIFF
--- a/dashboard-ui/src/components/SurveyResults/resultsTable.jsx
+++ b/dashboard-ui/src/components/SurveyResults/resultsTable.jsx
@@ -61,7 +61,7 @@ export function ResultsTable({ data }) {
             entryObj['End Time'] = new Date(entry?.timeComplete)?.toLocaleString();
             const timeDifSeconds = (new Date(entry?.timeComplete).getTime() - new Date(entry?.startTime).getTime()) / 1000;
             entryObj['Total Time'] = formatTime(timeDifSeconds);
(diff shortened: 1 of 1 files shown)


kaitlyn-sharo (approved):
Seems to work

nextcen-dgemoets (approved):
I started the dashboard on `main` and downloaded the Soartech Sub scenario delegation survey results.  I ran the deployment script and ran the dashboard under the `itm-395` branch and re-downloaded the Soartech Sub scenario delegation survey results.  The number of  filled in "What is your current role" values went from 3 to 15.

brianpippin (approved):
This seems to work but for the Dry Run Evaluation I think we need some major clean up of the survey json so this isn't as easily breakable, I think we should have a quick discussion about how to revamp the JSON some so that some of these long strings are not object properties and we use a consistent approach across, maybe like questionText, questionNumber, etc to organize it.  I would ask that we create a ticket to investigate it, but not start it as there is some higher priority work over top the Dashboard currently.

PR #70, opened 2024-05-29
Dev to Main

brianpippin:
Want to get the Dashboard updated to the latest before the PI Meeting.


dereknop (approved):

kaitlyn-sharo (commented):
Several columns on the http://localhost:3000/humanProbeData page are missing data, such as SimOrder and both KDMA columns (at the end), as well as additional columns in between. I don't know when this happened, but I made sure to run ingest and confirmed that Derek is seeing the same thing. Just wanted to mention it in case it is an unknown issue.

It doesn't look like any of this is part of this specific PR, but something to investigate and possibly fix? Everything else looks good to me.

kaitlyn-sharo (approved):

PR #71, opened 2024-06-14
Pull Survey Config from Mongo

kaitlyn-sharo:
Use with itm-464-mongo-config in the ingest repository. Follow the instructions in that PR to upload survey configs to mongo.

To test, run through the delegation survey. You can also change the survey version to 0.0 in the .env to run through without any required questions. This should be helpful for future development/testing.

Before running, you'll need to rerun the dashboard-graphql setup script to make sure the env gets captured as well as the new graphql queries.

Also check the survey results page. The delegation survey and survey results page are the only two that should have been updated by these changes.

Some of the code might look strange; due to the class-based components that we have to have because of SurveyJS, I had to do a few work-arounds to access the new redux store that was implemented (Derek and I thought the store may also be helpful for some of the upcoming tasks as well, so it won't just be used for the config). If you see something that looks really bad and you have a different suggestion, please feel free to mention it! I tried a lot of different methods here and this was the first that worked, so I stuck with it


dereknop:
Note to other reviewers that you'll have to rebuild the dashboard container as well as the graphql container since a couple of libraries were added to package.json.

dashboard-ui/src/components/App/index.jsx
```diff
@@ -144,134 +155,168 @@ export class App extends React.Component {
...
+                                return;
+                            }
+                            if (error) {
+                                console.warn("Error getting survey config: ", error);
+                                return;
+                            }
+                            // Get the survey configs one by one, since they are all needed for the results pages
+                            // and add the correct image urls back to them
+                            // then store them properly in localStorage
+                            for (const config of data.getAllSurveyConfigs) {
```
dereknop, line 171: This is a preference thing so feel free to ignore. But I'd probably move these for four loops into a separate function like this:
```
function matchImagesToConfig(data) {
    data.getAllSurveyConfigs.forEach(config => {
        config.survey.pages.forEach(page => {
            page.elements.forEach(el => {
                if (el.patients) {
                    el.patients.forEach(patient => {
                        const foundImg = data.getAllImageUrls.find(x => x._id === patient.imgUrl);
                        if (isDefined(foundImg)) {
                            patient.imgUrl = foundImg.url;
                        }
                    });
                }
            });
        });
        store.dispatch(addConfig({ id: config._id, data: config }));
    });
}
```
and then just call it from inside the query tag

kaitlyn-sharo: great call, this was all a result of experimentation and I never really considered how to make it cleaner. Done, but I changed the function name to setupConfigWithImages

dashboard-ui/src/components/Survey/survey.jsx
```diff
@@ -29,13 +29,22 @@ class SurveyPage extends Component {
             startTime: null,
             firstPageCompleted: false,
             surveyId: null,
-            surveyVersion: surveyConfig2x.version,
```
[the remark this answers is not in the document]
kaitlyn-sharo: Totally forgot to add this back in after hours of struggling with class based components! Thanks for the reminder, fixed now and I see the version in mongo when I test

dereknop (approved):

nextcen-dgemoets (commented):
Everything looks good.  I didn't really "review" the code because I'm scarcely qualified to do so.  But the dashboard itself looks fine.

PR #81, opened 2024-08-13
Text based UI improvements

dereknop:
Various UI tweaks for text-based scenarios. The image modal now behaves properly. Sometimes Adept was using character id's in their actions and sometimes using character names so I also added the character name when clicking on the image for a patient.


phile-caci (approved):
Didn't see any issues navigating around...

kaitlyn-sharo (approved):

PR #82, opened 2024-08-14
orderLog available in the excel download

dereknop:
Ewart needs to see the orderLog from the survey 3.0 results. 

To test, click through the survey (don't put 'test' in the participant ID) and then go to the survey results page. Click 'View Tabulated Data' and then make sure only survey version 3.0 is selected. Download the data to Excel and ensure you can see the orderLog field populated with an array of strings (should be 24 elements). This list will let Ewart know what order participants saw ADM/Scenario pairings along with additional info.

Note: The code for the survey results is getting pretty unmaintainable so this will be something to refactor once I get the chance.


phile-caci (approved):
LGTM!
Clicked through a survey, downloaded xlsx, has `Order Log` with 24 elements.

image

kaitlyn-sharo (approved):
Worked for me as well

PR #83, opened 2024-08-19
Itm 551

dereknop:
[ITM-551](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-551)

Requires running [Ingest pr](https://github.com/NextCenturyCorporation/itm-ingest/compare/ITM-551?expand=1) first.

New UI features
- More info modal for ST
- New colors for badges to indicate differences between normal and alarming vitals
- Label icons so you don't have to hover over them to see what they correspond to
- New colors for badges indicating the severity of injuries
- Vitals showing as unknown if assessing the character is an action in the action mapping
- Modal popup for transition scenes
- Patient images should be larger when you click on the magnifying glass in their container
- More info about the patient should be present (varies depending on ST vs. Adept)
- Supplies column should only take up necessary space (should no longer be growing and leaving white space if there are three patients)
- All CSS in separate file

[ITM-551]: https://nextcentury.atlassian.net/browse/ITM-551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dashboard-ui/src/components/TextBasedScenarios/medicalScenario.jsx
+++ b/dashboard-ui/src/components/TextBasedScenarios/medicalScenario.jsx
@@ -229,7 +229,7 @@ export class MedicalScenario extends SurveyQuestionElementBase {
         <Col md={9}>
         <Row>
             {this.patients.map((patient, index) => (
-              <Col md={5} key={index} className="mb-4">
+              <Col md={6} key={index} className="mb-4">
                 <Card className="h-100 border-0 shadow-sm">
                   <Card.Body style={{ overflow: 'visible' }}>
                     <Card.Title className="h4 mb-1">
--- a/dashboard-ui/src/components/TextBasedScenarios/medicalScenario.jsx
+++ b/dashboard-ui/src/components/TextBasedScenarios/medicalScenario.jsx
@@ -53,6 +53,14 @@ export class MedicalScenarioModel extends Question {
   set mission(mission) {
     this.setPropertyValue("mission", mission)
   }
+
+  get blockedVitals() {
+    return this.getPropertyValue("blockedVitals")
+  }
+
+  set blockedVitals(blockedVitals) {
+    this.setPropertyValue("blockedVitals", blockedVitals)
+  }
 }
 
 Serializer.addClass(
@@ -62,7 +70,8 @@ Serializer.addClass(
     { name: "supplies", default: [] },
     { name: "patients", default: [] },
     { name: "events", default: [] },
-    { name: 'mission', default: null}
+    { name: 'mission', default: null},
+    { name: 'blockedVitals', default: []}
   ],
   function () {
     return new MedicalScenarioModel("");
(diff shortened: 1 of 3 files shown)


kaitlyn-sharo (commented):
From a UX perspective, I think in the additional info section, those modals should not highlight on hover. It makes them look clickable, but they are not, which could be confusing to a user. 

Personally, I still think some of the text is too small, like on the Review Text Scenarios page now and some of the text on the additional info section.

I am loving the new colors for the vitals. However on my normal-sized laptop screen, Mental Status extends onto two lines which ruins the flow of the vitals section. I wonder if this can be fixed in any way? I don't know what the participants will be using to take the survey. 
image

Soartech: I find it strange that we can see the vitals of the patients when we choose who to treat, but then when we get brought to our first choice (assess/treat/etc), those vitals disappear. So now I could be remembering the vitals of the patient shown on page one, but I have no idea about the vitals of the new patient. Just a thought...this is a pretty rare occurrence anyways since most of the time ST does not offer the assess action.
Adept: Same thing, but more extreme. For example, in the desert, we are told all of the vitals during the first two probes, and then they switch to unknown when it asks who we want to assess. Should we be looking ahead for assess actions and not showing vitals until assess has occurred?

I still agree with Darren on the desert adept as well, that the US soldier should be shown during that transition scene, not sure what Adept said when you asked them.

Adept Desert: I "continued assessing the US soldier" and his vitals did not show up.

At the beginning of Adept shooter/victim I love the way you split up the answer options there. It's super clean. I don't think they're all split correctly though:
image
HOWEVER, I also recognize we are waiting for finalized adept scenarios and you are probably waiting for that rather than perfecting something that is not final yet

And finally, I'm still getting the pop up when I finish the scenario that says "are you sure?"

dereknop:
> At the beginning of Adept shooter/victim I love the way you split up the answer options there. It's super clean. I don't think they're all split correctly though: image HOWEVER, I also recognize we are waiting for finalized adept scenarios and you are probably waiting for that rather than perfecting something that is not final yet

Screenshot 2024-08-19 at 3 48 40 PM

Screenshot 2024-08-19 at 3 49 00 PM


That's how they have it set up in the yaml... I will need to ask them about this one because you are right this looks wrong. I believe the action conditions for those last three options in the first question are wrong

dereknop:
> From a UX perspective, I think in the additional info section, those modals should not highlight on hover. It makes them look clickable, but they are not, which could be confusing to a user.

Fair enough, I was getting carried away haha

> Personally, I still think some of the text is too small, like on the Review Text Scenarios page now and some of the text on the additional info section.


Okay I'll try making it a bit bigger


> I am loving the new colors for the vitals. However on my normal-sized laptop screen, Mental Status extends onto two lines which ruins the flow of the vitals section. I wonder if this can be fixed in any way? I don't know what the participants will be using to take the survey. 

I never noticed this because I always tested on a 2nd monitor. Good catch though. I'll try to make that container just a bit bigger

phile-caci (approved):
LGTM!

PR #88, opened 2024-09-03
Itm 570 delegation dre

kaitlyn-sharo:
Run the matching ingest PR first, then make sure you rerun the setup scripts, as the survey version in .env has changed. Check out the delegation survey review page.


phile-caci (approved):
Spot checked a few dozen surveys, looks good.  I think my only note would be that the font size is pretty small in the Medic Actions (not a deal breaker).

kaitlyn-sharo:
I agree with the font-size. I think that all has to do with the zoom change Derek made awhile ago. Everything has been pretty small on my screens since then but we've decided to leave it. I could probably increase the font size here to undo that change. Should be easy enough and I think it would make a huge difference

brianpippin (approved):
Looks good.  We just need to figure out what why some tad-baseline is missing but I clicked through and everything worked.

PR #89, opened 2024-09-03
Itm 572

dereknop:
[ITM-572](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-572)

First, run the [ingest pr](https://github.com/NextCenturyCorporation/itm-ingest/pull/30)

You should be able to see the `participantLog` collection in MongoDB. Use a PID from there for the text based scenarios and compare that the proper scenarios are shown in the proper order. The mappings for the `Text-1` and `Text-2` can be found on line 490 of `TextBasedScenariosPage.jsx`. After clicking through all of the scenarios, you should see separate documents added to the `userScenarioResults` collection for each scenario completed. 

Other UI things to test:
- Clicking 'M' on the end screen takes you back to the intro (allowing another participant to go through)
- If you enter a PID that does not map to any entry in the `participantLog`, you are prompted that this is the case (to ensure OSU is entering correct PID). If you don't proceed with the incorrect PID, you should just be able to re enter your information. If you do wish to proceed, a default config of the scenarios will be presented. Keep in mind that having 'test' in the PID will prevent any results from being uploaded. 

[ITM-572]: https://nextcentury.atlassian.net/browse/ITM-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

EDIT: Also, on the end screen confirm that the VR scenarios it tells the moderator to run match up to the ones in the participantLog for that participant


dereknop:
I forgot to mention that the graphql container will need to be rebuilt because I added a query to grab the participantLog

kaitlyn-sharo (approved):
Everything seemed to work as described

brianpippin (approved):
LGTM

PR #97, opened 2024-09-10
Text based refinement

dereknop:
No ingest script this time.

Changes:
- Running through the text-based scenario using a PID from the participant log will attach the order of the scenarios seen to the documents posted in Mongo. (i.e `scenarioOrder: ['AD-1', 'ST-2']`)
- internal injuries had a `location` of internal and a `name` of internal so it was printed as "Internal Internal" in the patient injuries section. It makes sense to include location if it is something like "Broken Right Arm" but I added in a check if the location is internal to not repeat the word. 
- Improved layout of patient cards in delegation survey. Before they were scrunched and the vitals were getting truncated on all of them. Should look a lot better now.
- Removed the 'no image available' container from Adept's training scenarios. If you look at IO1 and MJ1 on the review page you will see that the vitals just fill the space now.

### DIFF
--- a/dashboard-ui/src/components/Survey/dynamic.jsx
+++ b/dashboard-ui/src/components/Survey/dynamic.jsx
@@ -98,7 +98,7 @@ const Dynamic = ({ patients, situation, supplies, decision, dmName, actions, sce
         const patientCards = patients.map(patient => {
             if (visiblePatients[patient.name] && sceneCharacters.includes(patient.name)) {
                 return (
-                    <>{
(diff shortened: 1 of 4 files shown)


phile-caci (approved):
Looks good, tested the four changes and seems to be working.

nextcen-dgemoets (approved):
All of the changes look good.
The review modal part on comparison pages could be a bit wider, but everything listed here is fixed.
patient-card

brianpippin (approved):

PR #100, opened 2024-09-11
eval name and number

dereknop:
I've added evalName and evalNumber to Mongo documents. This will help with analyzing data further down the line. 

http://localhost:3000/text-based

You should have the following fields: 
Screenshot 2024-09-11 at 12 59 14 PM

### DIFF
--- a/dashboard-ui/src/components/Survey/survey.jsx
+++ b/dashboard-ui/src/components/Survey/survey.jsx
@@ -1142,6 +1142,11 @@ class SurveyPage extends Component {
             }
         }
 
+        if (this.state.surveyVersion == 4.0) {
+            this.surveyData['evalNumber'] = 4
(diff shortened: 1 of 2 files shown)


brianpippin:
I should probably ask this here as well, are we already doing this when saving the Delegation Survey as well?

phile-caci:
> I should probably ask this here as well, are we already doing this when saving the Delegation Survey as well?

We probably should?? It does have a survey_version, which is what we use for the drop down selection

brianpippin (approved):

nextcen-dgemoets (approved):
`evalName` and `evalNumber` appeared in `userScenarioResults` after completing the text scenarios.
`evalName` and `evalNumber` appeared in `surveyResults` after completing the delegation survey.

PR #101, opened 2024-09-11
url's need to include https

dereknop:


nextcen-dgemoets:
In the TA3 Server config, we use the same URLs, but we've been using them with `http://`.
Is this because the request is coming from a browser / React app and not an AWS server?

dereknop:
> In the TA3 Server config, we use the same URLs, but we've been using them with `http://`. Is https needed here because the request is coming from a browser / React app and not an AWS server?
Screenshot 2024-09-11 at 2 51 46 PM

I believe that is the case

nextcen-dgemoets (approved):

brianpippin (approved):

phile-caci (approved):
Works for me!

PR #102, opened 2024-09-11
account for MJ2 Gauze edge case

dereknop:
Scene 2A in Adept's MJ2 had to be tweaked to fit the text-based format. I had worked with Dustin to figure out how those mappings should work. I forgot to handle several probe responses to be sent from one answer (i.e 3 units of gauze for Babson and 2 for Alderson). This change allows for 5 probe responses to be sent off of selecting that answer.


nextcen-dgemoets (approved):
Tested locally with 202409113 MJ2 didn't see any errors in the console.
Then again, I didn't see any on `development` either.

PR #104, opened 2024-09-12
changed soartech adm loading logic

kaitlyn-sharo:
ST no longer cares about overlaps with adms that are in the same category as the baseline adm. I removed these from the col1to3 variable in order to free them to be used by the aligned and misaligned adms. 

https://nextcentury.atlassian.net/wiki/spaces/ITMC/pages/3213885465/Dry+Run+Evaluation+Delegation+Materials


phile-caci (approved):
Seems to be working, didnt see the removed targets

dereknop (approved):

PR #105, opened 2024-09-12
Survey debugging

dereknop:
Edge case Jennifer found.

If the scenario is IO5 or IO4, the adm belongs to parallax, and the alignment status is overlapped with the baseline, the alignment should always be recorded as misaligned (Was previously showing as aligned)

### DIFF
--- a/dashboard-ui/src/components/Survey/survey.jsx
+++ b/dashboard-ui/src/components/Survey/survey.jsx
@@ -1088,6 +1088,13 @@ class SurveyPage extends Component {
                     questions: {}
                 };
 
+                //comparison page
+                if (page?.name?.contains('vs')) {
+                    this.surveyData[pageName]['baselineName'] = page?.['baselineName']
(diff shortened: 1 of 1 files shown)


phile-caci (approved):
As far as I can tell the code looks good. Not getting any errors, but also didnt come across a IO4/5.

Is there a participantID that we can test with?

Also, since we're talking about IO4/5 do they need to be here?

image

nextcen-dgemoets:
@phile-caci Derek and I looked at this for a bit late this afternoon.  We used 202409142 since that's the one Jennifer used when she found the bug.  I removed all of the 202409142 entries from `userScenarioResults`, then ran through the text scenarios and tried to make my choices go right down the middle in my ADEPT choices (on Derek's advice).  Then I logged my `surveyConfigClone.pages` and sent Derek the results.  He wanted to see the expansion of Medic P21, so I obliged.

But around that time, there was more Teams discussion on the topic and ultimately Derek suggested we stick a pin in it and revisit tomorrow/Friday.  So that's what we did.  :)

PR #109, opened 2024-09-13
Dev to Main

brianpippin:


nextcen-dgemoets (commented):
I'm still getting weird stuff that nobody else is getting presumably because of something with my DB, so I'll recuse myself from this review.

Once dev is pushed to main and deployed, I'll double-check that the things I was seeing are gone.

dereknop (approved):

phile-caci (approved):
Looks good, other than:

```
Uncaught TypeError: mission.medical_policies.map is not a function
    MoreDetailsModal moreDetailsModal.jsx:44
    React 17
    handleConfigSelect ReviewTextBased.jsx:75
    onClick ReviewTextBased.jsx:128
    React 15
moreDetailsModal.jsx:44
    MoreDetailsModal moreDetailsModal.jsx:44
    React 17
    handleConfigSelect ReviewTextBased.jsx:75
    onClick ReviewTextBased.jsx:128
    React 15

```

which will be addressed post DRE

nextcen-dgemoets (approved):
My DB got fixed (forgot to run a script) and things look good.
Derek even tracked down my stacktrace I was getting:
  - when a PID is not in the participant list, but is used then reused.

PR #113, opened 2024-09-19
added randomization and recorded order in delegation survey

kaitlyn-sharo:
We need to randomize baseline/aligned/misaligned in the delegation survey. It would also be helpful to have a record of the order of pages shown. This ticket handles that, adding randomization for each set of ADMs within a TA2 and scenario, and recording the order of pages in orderLog in mongo.

To test, start the survey and make sure mongo records the orderLog properly. 

Also make sure that the order of pages is random. To quickly see this, if you look at the comparison page, you'll see M1 vs M2 vs M3, where M1 is baseline, M2 is aligned, and M3 is misaligned. Make sure that the individual medic pages do not always show up as M1, M2, M3, but sometimes M2, M3, M1, or M3, M1, M2, etc


brianpippin (approved):
Worked for me.

PR #116, opened 2024-09-20
added RQ variable definitions

kaitlyn-sharo:
For each RQ page, look for the View Variable Definitions under each table. I would recommend downloading the pdf, making sure that the pdf only contains the information for that specific table, then comparing the information in the website table to the pdf.


dereknop (approved):
Downloading PDF worked for me and it appears to match up with the table in the dashboard. One nitpicky thing I would suggest is putting '1-4:' on it's own line since it is separate from the order table in the pdf.
Screenshot 2024-09-23 at 9 11 20 AM
Screenshot 2024-09-23 at 9 11 27 AM

nextcen-dgemoets (approved):
I couldn't find any problems or inconsistencies.

PR #118, opened 2024-09-23
Generalize survey results (4.0)

kaitlyn-sharo:
In the survey results page, once you select a scenario from survey version 4, you should see a generalize toggle. Toggling it on and off will switch between aligned/misaligned/baseline and the specific medics the questions were answered for.


brianpippin (approved):
This looks great, I think Jennifer is going to love this view for the DRE!

nextcen-dgemoets:
Maybe this is just my platform/browser, but the new control doesn't look like a toggle button.  And maybe there's nothing we can do about it as it's just a toolkit thing, but the toggle just looks like a blue dot:
generalize
It sort of looks like a toggle when it's toggled off:
generalize-off

Also, maybe this is just my lack of sophistication regarding what these charts are showing, but what does "generalize" mean in this context?  The PR says, "switch between aligned/misaligned/baseline and the specific medics the questions were answered for"  So does that mean that it's a toggle to present the data by medic vs. by ADM-- a different way to slice and dice the data?  Is it zooming out on the data?  I'm not sure.  Could we either have a tooltip or a description somewhere, or label the on and off states to explain what they are?  Again, I admit maybe this is just a "me problem" and it will be obvious to anybody using this what a "generalize data" toggle will mean/do/be.

kaitlyn-sharo:
> Also, maybe this is just my lack of sophistication regarding what these charts are showing, but what does "generalize" mean in this context? The PR says, "switch between aligned/misaligned/baseline and the specific medics the questions were answered for" So does that mean that it's a toggle to present the data by medic vs. by ADM-- a different way to slice and dice the data? Is it zooming out on the data? I'm not sure. Could we either have a tooltip or a description somewhere, or label the on and off states to explain what they are? Again, I admit maybe this is just a "me problem" and it will be obvious to anybody using this what a "generalize data" toggle will mean/do/be.

Instead of showing the data medic-by-medic, like in ITM-585 (or with generalize data toggled off), which will usually only show one or two responses per medic and will show a lot of graphs, this will combine several medics into one graph based on whether that ADM was aligned/misaligned/baseline for the participant. So one participant may have had Medic A as their Parallax Aligned medic and one may have had Medic D. The original view (generalize data toggled off) will show Medic A and Medic D separately. The generalized view will combine Medic A and Medic D into one graph under the title Parallax Aligned, because really that's the data we want to see, is how did people answer based on the alignment of the ADM to them, not how did everyone answer for Medic A, Medic B, Medic C, regardless of the alignment and targets. 

I am 100% open to changing the text if Generalize Data is not clear. It was the first thing that popped into my head and I did not think much about it. I was more concerned with getting the page working :)

nextcen-dgemoets:
For "Generalize Data", how about making it a radio button with choices of, say, "View by Medic" and "Aggregate by ADM"?

Or, having a label "View By" then a radio button "Medic" and "ADM".  Making it a radio button will also address the toggle visual issue (it still looks weird for me on both Chrome and FF).

Or, if the generalized view is really the data we want to see, do we need the "un-generalized" view at all?

dereknop (approved):

nextcen-dgemoets (approved):
Yay-- disabling clicking on the bars really helps; thanks @dereknop !

We can revisit Generalize after we know what view(s) Jennifer wants.

kaitlyn-sharo:
> Yay-- disabling clicking on the bars really helps; thanks @dereknop !
> 
> We can revisit Generalize after we know what view(s) Jennifer wants.

I have Jennifer's feedback on this and will be implementing by the end of the day

PR #119, opened 2024-09-24
Itm 601 sim play-by-play

kaitlyn-sharo:
Updated http://localhost:3000/human-results for DRE.

Organizes DRE scenarios by scenario rather than environment, taking away the toggle between Adept/Soartech/Freeform, since that is no longer relevant. To see all probes, run the latest version of the probe_matcher, as the manually entered probes (dragging probes for QOL/VOL scenarios) have been put in the database in the latest commits [here](https://github.com/NextCenturyCorporation/itm-ingest/pull/40). 

Make sure switching between evaluations, scenarios, and participants works smoothly, and that the data looks appropriate. I fixed a typo in our graphql calls, so you will need to rerun the startup script. 

This also includes a minor fix so that entering http://localhost:3000/humanSimParticipant does not cause an error.

Note: built off of generalize-survey-results AND derek's branch to include all recent changes. That's why it shows extra changes that are not part of this branch.

### DIFF
--- a/dashboard-ui/src/components/AggregateResults/DataFunctions.js
+++ b/dashboard-ui/src/components/AggregateResults/DataFunctions.js
@@ -467,52 +467,54 @@ function getAttributeAlignment(res, att, medics, q, trans) {
     return { 'align': alignTally > 0 ? align / alignTally : null, 'misalign': misalignTally > 0 ? misalign / misalignTally : null };
 }
 
-function populateHumanDataRow(rowObject) {
+function populateHumanDataRow(rowObject, version) {
     let adept = 0;
     let soartech = 1;
     if(rowObject[0]._id.indexOf("_st_") > -1) {
         adept = 1;
         soartech = 0;
     }
+    let returnObj = {};
+    if (version == 3) {
+        returnObj = {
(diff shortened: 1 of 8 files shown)


nextcen-dgemoets (commented):
UI looks really good.  The phrase, "Please select an environment and participant to view results" is MRE-centric (as opposed to "scenario and participant"), but I don't think this will cause user confusion.

One data question, though-- is it expected that none of the QOL scenarios have probe 12 (aka `qol-dre-3-eval-Probe-12`) answered by any of the participants?

brianpippin:
> UI looks really good. The phrase, "Please select an environment and participant to view results" is MRE-centric (as opposed to "scenario and participant"), but I don't think this will cause user confusion.
> 
> One data question, though-- is it expected that none of the QOL scenarios have probe 12 (aka `qol-dre-3-eval-Probe-12`) answered by any of the participants?

Should be fairly easy to change this text if evalNumber = 4 in the UI.

nextcen-dgemoets (approved):
Re-ran the probe matcher and probe 12 is there.
UI looks good to me!

dereknop (approved):

PR #120, opened 2024-09-25
ITM-565 - Updated SSL Certificate

phile-caci:
This is already running on Prod (it's the only way to test)

You wont see the certificates committed into the repo.
Certs were installed on our nginx **and** CISTAC needs to install the same certs on CACI's F5 load balancer.

If you go to the UI website, you should see the new SSL (on the address bar, to the left of the url, should be a lock icon, you should be able to navigate to the certificate).  You may need to do a hard refresh of the website (cntrl+shift+R on windows or Cmd+Opt+R on mac)

The new expiration date: Saturday, August 23, 2025 at 2:48:49 PM

I'll post a write-up on Confluence for the install.

### DIFF
--- a/docker_setup/docker-compose.yml
+++ b/docker_setup/docker-compose.yml
@@ -10,6 +10,7 @@ services:
     volumes:
       - ./resources/nginx.conf:/etc/nginx/nginx.conf
       - ./certs:/certs
+      - ./nginxpwd:/nginxpwd
     ports:
       - "80:80"
       - "443:443"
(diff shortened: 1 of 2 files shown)


nextcen-dgemoets (approved):
Cert looks good on Chrome and FF-- expiration date is 8/23/25.
Can access `https://darpaitm.caci.com/login` and `https://darpaitm.caci.com/adept/api/v1/alignment_target_ids` from non-CACI device, and it shows the valid cert.

brianpippin (approved):
I can see the new cert!

PR #122, opened 2024-09-26
Itm 603 human probe data

kaitlyn-sharo:
NOTE: This will not work without running the probe matcher, same as the play-by-play page, because the data for this page is only populated in the database by the probe matcher. Please do not expect this to work on production (or locally) if the probe matcher has not been run!

This updates http://localhost:3000/humanProbeData for DRE. Instead of organizing by environment, it organizes the data by Adept scenario, since we need different headers depending on the scenario. The values in the cells are KDMA values. If a cell is empty (no '-'), it means that participant did not encounter that probe in the sim. If a cell has a '-', it means they encountered the probe, but a KDMA does not exist for the probe. 

Please make sure downloads for MRE and DRE both work. The DRE download has 3 worksheets, one for each Adept scenario, due to the different headers


dereknop (approved):
Looks good and the downloads worked for me

nextcen-dgemoets:
So the data and the UI all looks good.  Both downloads looked just like the pages (from a data perspective).

The only thing I'm not following is how environment (MRE) turned into ADEPT scenario (DRE).  Why does a table with a header of an ADEPT scenario (DryRunEval-MJ4-eval) have ST stuff in it (QOL_1, VOL_6, etc.).  I'm guessing that maybe in the sim, the scenarios are combined / concatenated such that the participant smoothly transitions from, say, `DryRunEval-MJ4-eval` to `qol-dre-1-eval`.  Aren't all ST scenarios urban (at least in the YAML)?

One other small thing-- could `mre-` or `dre-` be prepended to the saved filenames, e.g., `dre_human_sim_data.xlsx`?

kaitlyn-sharo:
> So the data and the UI all looks good. Both downloads looked just like the pages (from a data perspective).
> 
> The only thing I'm not following is how environment (MRE) turned into ADEPT scenario (DRE). Why does a table with a header of an ADEPT scenario (DryRunEval-MJ4-eval) have ST stuff in it (QOL_1, VOL_6, etc.). I'm guessing that maybe in the sim, the scenarios are combined / concatenated such that the participant smoothly transitions from, say, `DryRunEval-MJ4-eval` to `qol-dre-1-eval`. Aren't all ST scenarios urban (at least in the YAML)?
> 
> One other small thing-- could `mre-` or `dre-` be prepended to the saved filenames, e.g., `dre_human_sim_data.xlsx`?

I can prepend mre or dre, that's easy.

I agree that the organization is strange. It couldn't be by environment for a few reasons. 1: all ST scenarios are urban; 2: Jennifer wanted one row per participant this time, so we needed to combine ST and Adept. As I explained in the description, each adept scenario has a different header due to different probe names and a different number of probes, whereas ST is straightforward and stays the same from one to the next. The other possibility was doing every combination of Adept and ST, such as 1/1, 1/2, 1/3, 2/1, 2/2, 2/3, 3/1, 3/2, 3/3, but that's a lot of combinations with less data. I think they like seeing as much data concatenated as possible. I did run this by Brian who said it's okay for now and we will wait for Jennifer's feedback on the organization. Though again I'll reiterate that I agree this is weird, but I think it's the cleanest, most straightforward way to combine the data until Jennifer says otherwise

nextcen-dgemoets:
> I agree that the organization is strange. It couldn't be by environment for a few reasons. 1: all ST scenarios are urban; 2: Jennifer wanted one row per participant this time, so we needed to combine ST and Adept. As I explained in the description, each adept scenario has a different header due to different probe names and a different number of probes, whereas ST is straightforward and stays the same from one to the next. The other possibility was doing every combination of Adept and ST, such as 1/1, 1/2, 1/3, 2/1, 2/2, 2/3, 3/1, 3/2, 3/3, but that's a lot of combinations with less data. I think they like seeing as much data concatenated as possible. I did run this by Brian who said it's okay for now and we will wait for Jennifer's feedback on the organization. Though again I'll reiterate that I agree this is weird, but I think it's the cleanest, most straightforward way to combine the data until Jennifer says otherwise

Makes sense-- I thought it was something like this.  So, it looks like the first set only involves `DryRunEval-MJ4-eval` (hence its name/title), plus ST1 and ST2.  And the first row (PID `20249201`) is `DryRunEval-MJ4-eval` + `qol-dre-2-eval` + `vol-dre-2-eval`.  And row 2 (PID `20249206`) is `DryRunEval-MJ4-eval` + `qol-dre-1-eval` + `vol-dre-1-eval`.  Is this right?

If so, yeah, I'm not sure if I have a better suggestion.  :)

kaitlyn-sharo:
> Makes sense-- I thought it was something like this. So, it looks like the first set only involves `DryRunEval-MJ4-eval` (hence its name/title), plus ST1 and ST2. And the first row (PID `20249201`) is `DryRunEval-MJ4-eval` + `qol-dre-2-eval` + `vol-dre-2-eval`. And row 2 (PID `20249206`) is `DryRunEval-MJ4-eval` + `qol-dre-1-eval` + `vol-dre-1-eval`. Is this right?
> 
> If so, yeah, I'm not sure if I have a better suggestion. :)

Not fully. Each set involves the Adept scenario in the title and whatever ST VOL/QOL scenarios the participants did. So as more participants take it, we might find that all 3 VOL/QOL scenarios show up. But aside from that, yes. It's just the limited amount of participants that is causing only the first two ST scenarios to show up right now

nextcen-dgemoets (approved):
Prepending of `mre_` or `dre_` happened in `ITM-600-aggregate-page` branch, so approving this one.

PR #124, opened 2024-09-27
Session ids

kaitlyn-sharo:
The server session ids used to get alignment for text scenarios have been added to the RQ1-3 graph.

The server session ids used by the adms to get alignment have been added to the RQ 22-23 graph.

The server session ids used to get alignment for the sim have been added to the human sim probe graphs. You will need to run the latest probe matcher with RUN_ALIGNMENT=True to see these values. 

Must run docker script due to a minor change in one of the queries

### DIFF
--- a/dashboard-ui/src/components/Account/adminPage.jsx
+++ b/dashboard-ui/src/components/Account/adminPage.jsx
@@ -128,7 +128,6 @@ function EvaluationIDSTable({data}){
         }else{
             setSelectedEvals(selectedEvals.filter(id=>id !== checkedId))
         }
-        console.log(selectedEvals);
     }
     
     return(
--- a/dashboard-ui/src/components/AggregateResults/DataFunctions.js
+++ b/dashboard-ui/src/components/AggregateResults/DataFunctions.js
@@ -518,6 +518,82 @@ function populateHumanDataRow(rowObject, version) {
             }
         }
     }
+    else if (version == 4) {
+        returnObj = {
+            "Participant": rowObject[0].pid
+        };
+        const adept_mapping = { 'MJ2': 1, 'MJ4': 2, 'MJ5': 3 };
+        if (rowObject[0].scenario_id.includes('ol')) {
+            returnObj['ST_Scenario'] = rowObject[0].scenario_id.split('ol-dre-')[1].split('-')[0];
+            returnObj[rowObject[0].scenario_id.split('ol')[0] == 'q' ? 'QOL_Session_Id' : 'VOL_Session_Id'] = rowObject[0].data.alignment['sid'];
+        }
+        else {
+            returnObj['ADEPT_Scenario'] = adept_mapping[rowObject[0].scenario_id.split('-')[1].split('-')[0]];
+            returnObj['ADEPT_Session_Id'] = rowObject[0].data.alignment['sid'].replaceAll('"', '');
+        }
+
+        const mj2ProbeIds = ["Probe 2", "Probe 2A-1", "Probe 2B-1", "Probe 3-B.2", "Probe 4", "Probe 4-B.1", "Probe 4-B.1-B.1", "Probe 5", "Probe 5-A.1", "Probe 5-B.1", "Probe 6", "Probe 7", "Probe 8", "Probe 9", "Probe 9-A.1", "Probe 9-B.1", "Probe 10"];
+        const mj4ProbeIds = ["Probe 1", "Probe 2 kicker", "Probe 2 passerby", "Probe 2-A.1", "Probe 2-D.1", "Probe 2-D.1-B.1", "Probe 3", "Probe 3-A.1", "Probe 3-B.1", "Probe 6", "Probe 7", "Probe 8", "Probe 9", "Probe 10", "Probe 10-A.1", "Probe 10-A.1-B.1", "Probe 10-B.1", "Probe 10-C.1"];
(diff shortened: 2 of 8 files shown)


nextcen-dgemoets:
This might have already been there and I didn't notice it (in other words, not part of this ticket), but in RQ1 Data's chart, the "Delegation preference" columns (both A/B and A/M) have some values of y and n.  Maybe this has to do with the Note in the definition:  "Note: response backfilled to ADMs in the same block to aid in calculation of chosen or not chosen RQ3.2"  But it's a little weird to see y and n when you're expecting A and B.  I imagine there's nothing to be done.

kaitlyn-sharo:
> This might have already been there and I didn't notice it (in other words, not part of this ticket), but in RQ1 Data's chart, the "Delegation preference" columns (both A/B and A/M) have some values of y and n. Maybe this has to do with the Note in the definition: "Note: response backfilled to ADMs in the same block to aid in calculation of chosen or not chosen RQ3.2" But it's a little weird to see y and n when you're expecting A and B. I imagine there's nothing to be done.

Jennifer said to use y and n, as in "y": yes, this ADM was chosen during the delegation or "n": no, this ADM was not chosen during the delgation

nextcen-dgemoets:
More relevant to this ticket, perhaps, is that there are no session_ids when the `ADM_Type` is "comparison".  I imagine this is expected/explainable.  The data dictionary ("Definitions" table) has a blank definition for `ADM_Type`, but says the levels (values?) are Baseline or Aligned.

kaitlyn-sharo:
> Slightly more relevant to this ticket, perhaps, is that there are no session_ids when the `ADM_Type` is "comparison". I imagine this is expected/explainable. The data dictionary ("Definitions" table) has a blank definition for `ADM_Type`, but says the levels (values?) are Baseline or Aligned.

Also something Jennifer said to do. She had considered having "comparison" as the adm type for comparison pages, but that's not an adm type, it's a page type, so we leave it blank because it's baseline and aligned being compared

nextcen-dgemoets:
The various session IDs should be added to the various pop-up Definitions tables.  Several of these tables are also missing other fields-- especially RQ2.2 & 2.3 Data-- some of which would explain what's going on with these tables.  But other than the session IDs, it's not specifically about this ticket.

kaitlyn-sharo:
> The various session IDs should be added to the various pop-up Definitions tables. Several of these tables are also missing other fields-- especially RQ2.2 & 2.3 Data-- some of which would explain what's going on with these tables. But other than the session IDs, it's not specifically about this ticket.

I already made a ticket for this. I cannot change it until Jennifer updates the pdfs. What other fields are missing?

nextcen-dgemoets:
>  What other fields are missing?

At least on my local version, RQ1 and RQ3's Variable Definitions table has empty definitions for `Delegator_grp`, `Attribute`, `TA2_Name`, and `ADM_Type`.  Also, `Server Session ID (Delegator)` isn't in these tables.  It may be obvious what many of these are, but it's nice to fill it out for completeness, IMO.

In RQ2.2 & 2.3's Variable Definitions table, every definition is blank except for `Trial_ID`, and there is no column for the new `Aligned Server Session ID` and `Baseline Server Session ID`.

kaitlyn-sharo:
> > What other fields are missing?
> 
> At least on my local version, RQ1 and RQ3's Variable Definitions table has empty definitions for `Delegator_grp`, `Attribute`, `TA2_Name`, and `ADM_Type`. Also, `Server Session ID (Delegator)` isn't in these tables. It may be obvious what many of these are, but it's nice to fill it out for completeness, IMO.
> 
> In RQ2.2 & 2.3's Variable Definitions table, every definition is blank except for `Trial_ID`, and there is no column for the new `Aligned Server Session ID` and `Baseline Server Session ID`.

As we discussed on Teams, as long as this matches the downloads that Jennifer created (which in the other reviews, everyone confirmed that, and nothing has changed since then), there is no action to take here

nextcen-dgemoets:
> > At least on my local version, RQ1 and RQ3's Variable Definitions table has empty definitions for `Delegator_grp`, `Attribute`, `TA2_Name`, and `ADM_Type`. Also, `Server Session ID (Delegator)` isn't in these tables. It may be obvious what many of these are, but it's nice to fill it out for completeness, IMO.
> > In RQ2.2 & 2.3's Variable Definitions table, every definition is blank except for `Trial_ID`, and there is no column for the new `Aligned Server Session ID` and `Baseline Server Session ID`.
> 
> As we discussed on Teams, as long as this matches the downloads that Jennifer created (which in the other reviews, everyone confirmed that, and nothing has changed since then), there is no action to take here

What about adding `Server Session ID (Delegator)` to the table, which was added for this PR--  even if we don't have a definition yet?  Or is that covered in some other ticket?  At a basic level, it's like updating the docs for the changes covered in a PR, but as long as it's covered in some other ticket, that's fine.

kaitlyn-sharo:
> > > At least on my local version, RQ1 and RQ3's Variable Definitions table has empty definitions for `Delegator_grp`, `Attribute`, `TA2_Name`, and `ADM_Type`. Also, `Server Session ID (Delegator)` isn't in these tables. It may be obvious what many of these are, but it's nice to fill it out for completeness, IMO.
> > > In RQ2.2 & 2.3's Variable Definitions table, every definition is blank except for `Trial_ID`, and there is no column for the new `Aligned Server Session ID` and `Baseline Server Session ID`.
> > 
> > 
> > As we discussed on Teams, as long as this matches the downloads that Jennifer created (which in the other reviews, everyone confirmed that, and nothing has changed since then), there is no action to take here
> 
> What about adding `Server Session ID (Delegator)` to the table, which was added for this PR-- even if we don't have a definition yet? Or is that covered in some other ticket? At a basic level, it's like updating the docs for the changes covered in a PR, but as long as it's covered in some other ticket, that's fine.

As I mentioned before, I have a ticket for updating the variables for these graphs.

nextcen-dgemoets:
I see.  I thought the ticket was for the things that don't specifically pertain to this ticket.

nextcen-dgemoets (approved):

dereknop (approved):
Nice job on these tables, I see the server session ids populating

PR #125, opened 2024-09-27
Itm 600 aggregate page

kaitlyn-sharo:
Made the DRE version of http://localhost:3000/humanSimParticipant 

Now labelled "Participant-Level Data" for DRE. 

Variable Definitions will be fixed in a separate ticket. 

Delegation is not filled out because I am not sure how to translate it from what we did in MRE to DRE yet. Waiting on an answer from Jennifer. I will put that in the Variable Definitions ticket. 

PostVRState is not filled out because of the missing data that has now been fixed (data for that page will now be collected).

Graphs have been added for ST's KDEs.


dashboard-ui/src/components/AggregateResults/DataFunctions.js
```diff
@@ -6,7 +6,16 @@ const SIM_MAP = {
...
+    "desert": 4,
+    "dryrun-adept-eval-MJ2": 1,
+    "dryrun-adept-eval-MJ4": 2,
+    "dryrun-adept-eval-MJ5": 3,
+    "dryrun-soartech-eval-qol1": 1,
+    "dryrun-soartech-eval-qol2": 2,
+    "dryrun-soartech-eval-qol3": 3,
+    "dryrun-soartech-eval-vol1": 1,
+    "dryrun-soartech-eval-vol2": 2,
+    "dryrun-soartech-eval-vol3": 3,
```
nextcen-dgemoets, line 18: Looks like an extra comma.

nextcen-dgemoets:
Things look good.  In some cases (e.g., pid=20249214, QOL_KDMA_Text), when I clicked "View Graph" it just had some tiny text that said, "Internal Server Error".  Based on the console output, it seems to happen when we call the soartech API with an unknown session_id.  That row also has no `MJ_KDMA_Text` or `IO_KDMA_Text`, although it does have `MJ_KDMA_Sim` and `IO_KDMA_Sim`.  Maybe somebody completed the sim part but didn't to the text scenarios?  Hmm, yes probably, because e.g. `AD_Scenario_Text` and `QOL_Scenario_Text` are also blank.  Not sure if other folks have this row in their DB.

I guess one possible suggestion is that when the `session_id` is blank or unknown, we just put a `-` instead `View Graph` that shows an error in the pop-up.

kaitlyn-sharo:
> Things look good. In some cases (e.g., pid=20249214, QOL_KDMA_Text), when I clicked "View Graph" it just had some tiny text that said, "Internal Server Error". Based on the console output, it seems to happen when we call the soartech API with an unknown session_id. That row also has no `MJ_KDMA_Text` or `IO_KDMA_Text`, although it does have `MJ_KDMA_Sim` and `IO_KDMA_Sim`. Maybe somebody completed the sim part but didn't to the text scenarios? Hmm, yes probably, because e.g. `AD_Scenario_Text` and `QOL_Scenario_Text` are also blank. Not sure if other folks have this row in their DB.
> 
> I guess one possible suggestion is that when the `session_id` is blank or unknown, we just put a `-` instead `View Graph` that shows an error in the pop-up.

That guy ruins everything. For some reason I thought this would happen automatically, but I've got the code in there now to fix it.

nextcen-dgemoets:
> That guy ruins everything. For some reason I thought this would happen automatically, but I've got the code in there now to fix it.

Looks great!  Do you know what happens if `darpaitm.caci.com` is down?  [Not that that would *ever* happen...]
I mean, we can't expect miracles, but it would be nice if it didn't crash or unduly hang the dashboard.

kaitlyn-sharo:
> > That guy ruins everything. For some reason I thought this would happen automatically, but I've got the code in there now to fix it.
> 
> Looks great! Do you know what happens if `darpaitm.caci.com` is down? [Not that that would _ever_ happen...] I mean, we can't expect miracles, but it would be nice if it didn't crash or unduly hang the dashboard.

If darpaitm.caci.com is down, the only two things that should be affected are the text scenarios, because at the end the alignment scores will not be able to be gotten from the servers, so we'd have to make an ingest script to fill that in after, of course that means that the delegation survey wouldn't work properly because the alignment scores aren't in to choose the delegators. 

As for the review pages, all it should do here is say Internal Server Error or Page Not Found or something like that when a "view graph" link is chosen.

nextcen-dgemoets:
> As for the review pages, all it should do here is say Internal Server Error or Page Not Found or something like that when a "view graph" link is chosen. If you wanted to test this, you could change .env to something nonsensical, i.e. add a typo in the ST url and I believe this would be the behavior seen...Remember when you change the .env you have to rerun the startup script to enact changes

Yep, the pop-up gives the frown face and, "darpaitm-foobar.caci.com’s server IP address could not be found."  It's possible it would give another message (like a timeout) if the address was valid but didn't respond, but either way, things look good.

nextcen-dgemoets (approved):
Looks great!

dereknop (approved):
Nice, ran ingest and see the graphs populating. Didn't encounter any errors opening them up

PR #126, opened 2024-09-27
Toggle survey version

dereknop:
[ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-518)

[Ingest pr](https://github.com/NextCenturyCorporation/itm-ingest/pull/46) first 

http://localhost:3000/admin

Verify that you can select a different survey version in the drop-down. Then http://localhost:3000/survey to see that the survey config has changed. 

Added some CSS to the admin page

Note: you must rebuild your docker container because a new query and mutation were added.

### DIFF
--- a/dashboard-ui/src/components/Account/adminPage.jsx
+++ b/dashboard-ui/src/components/Account/adminPage.jsx
@@ -239,10 +239,8 @@ function AdminPage({ currentUser }) {
     const [getConfigs, { loading: configsLoading, error: configsError, data: configsData }] = useLazyQuery(GET_CONFIGS, {
         fetchPolicy: 'no-cache',
         onCompleted: async (data) => {
-            console.log("GetConfigs completed:", data);
             try {
                 await setupConfigWithImages(data);
-                console.log(data.getAllSurveyConfigs)
             } finally {
                 setIsLoading(false);
             }
--- a/dashboard-ui/src/components/Account/adminPage.jsx
+++ b/dashboard-ui/src/components/Account/adminPage.jsx
@@ -1,36 +1,54 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { withRouter } from 'react-router-dom';
-import { Query } from 'react-apollo';
+import { Query, useQuery, useMutation } from 'react-apollo';
 import gql from 'graphql-tag';
 import DualListBox from 'react-dual-listbox';
-import { useMutation } from 'react-apollo';
-import { Button, Modal } from 'react-bootstrap';
+import { Button, Modal, Form } from 'react-bootstrap';
 import '../../css/admin-page.css';
+import { useSelector } from "react-redux";
+
 const getUsersQueryName = "getUsers";
 const GET_USERS = gql`
-    query getUsers{
+    query getUsers {
         getUsers
-    }`;
+    }
+`;
 
 const GET_EVAL_IDS = gql`
-    query getEvalIds{
+    query getEvalIds {
         getEvalIds
-    }`;
+    }
+`;
 
 const UPDATE_EVAL_IDS_BYPAGE = gql`
-    mutation updateEvalIdsByPage($evalNumber: Int!, $field: String!, $value: Boolean!){
+    mutation updateEvalIdsByPage($evalNumber: Int!, $field: String!, $value: Boolean!) {
         updateEvalIdsByPage(evalNumber: $evalNumber, field: $field, value: $value) 
-    }`;
+    }
+`;
 
 const UPDATE_ADMIN_USER = gql`
-    mutation updateAdminUser($username: String!, $isAdmin: Boolean!){
+    mutation updateAdminUser($username: String!, $isAdmin: Boolean!) {
         updateAdminUser(username: $username, isAdmin: $isAdmin) 
-    }`;
+    }
+`;
 
 const UPDATE_EVALUATOR_USER = gql`
-    mutation updateEvaluatorUser($username: String!, $isEvaluator: Boolean!){
+    mutation updateEvaluatorUser($username: String!, $isEvaluator: Boolean!) {
         updateEvaluatorUser(username: $username, isEvaluator: $isEvaluator) 
-    }`;
+    }
+`;
+
+const GET_CURRENT_SURVEY_VERSION = gql`
+    query getCurrentSurveyVersion {
+        getCurrentSurveyVersion
+    }
+`;
+
+const UPDATE_SURVEY_VERSION = gql`
+    mutation updateSurveyVersion($version: String!) {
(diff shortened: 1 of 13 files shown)


nextcen-dgemoets:
- If I change from version 4 to version 3, then confirm the popup, then go Home, then go back to Admin, it shows my current version is still 4.  I didn't see a save button or anything, so I assume it was supposed to happen.  In general, it wasn't showing what I thought was the current survey version based on my previous selection (with Confirm).
- If my version is 4 and I select version of 4 in the dropdown, it still shows the confirm pop-up.
- I selected version 0 or 1 from the dropdown, then http://localhost:3000/survey gave me a stacktrace.  Do we need optional chaining somewhere?

kaitlyn-sharo:
> * If I change from version 4 to version 3, then confirm the popup, then go Home, then go back to Admin, it shows my current version is still 4.  I didn't see a save button or anything, so I assume it was supposed to happen.  In general, it wasn't showing what I thought was the current survey version based on my previous selection (with Confirm).
> * If my version is 4 and I select version of 4 in the dropdown, it still shows the confirm pop-up.
> * I selected version 0 or 1 from the dropdown, then http://localhost:3000/survey gave me a stacktrace.  Do we need optional chaining somewhere?

Just to build off of Darren's, I see the same behavior where the version will not show the version I chose if I return to the page unless I refresh. So when I first change the version, it shows correctly, and the survey I expect to show up does in the survey, but returning to the admin page will revert the shown choice (just on the admin page)

The same thing happens for me with the confirm, but I wouldn't consider it something that needs to be changed.

Version 0 gives me a stack trace, verison 1 does not.

kaitlyn-sharo:
Just doing some investigation on the stack trace. It looks like it's happening because of the changes that were made to update_survey_config in the ingest repo. There's a comment that says "not using for 7-16" that would have created the 'validSingleSets' collection. Instead, it ends up empty. My guess is this is the problem. I was not part of that, because I was in Hawaii, so I don't know if we want to change it back for versions 0 and 1, or if we need to change the dashboard accordingly.

dereknop:
> Just doing some investigation on the stack trace. It looks like it's happening because of the changes that were made to update_survey_config in the ingest repo. There's a comment that says "not using for 7-16" that would have created the 'validSingleSets' collection. Instead, it ends up empty. My guess is this is the problem. I was not part of that, because I was in Hawaii, so I don't know if we want to change it back for versions 0 and 1, or if we need to change the dashboard accordingly.

I believe this was a change for survey version 2.1 I want to say... can't recall. Either way I'll get to the bottom of it

dereknop:
Thanks for the feedback. I'll remove the current survey version as an option to address your point about the confirmation showing unnecessarily @nextcen-dgemoets . Also, I believe the other issue you guys pointed out is a caching issue so I think I know a fix for that.

kaitlyn-sharo:
Oh I also wanted to mention I think the admin page looks great. I just got distracted by darren's comment. nice css work

nextcen-dgemoets:
Not sure how I managed to get to Survey Version 0, but the DB agrees with the screen here:  version 0.  I'll see if I can duplicate.
survey-0

nextcen-dgemoets:
Deleted previous comment-- I forgot that Version 0 is a valid version number!

nextcen-dgemoets (approved):
The fixes look good to me.  There are a couple of console.logs to remove (`dynamic.jsx` and `supplies.jsx`).

I wonder if we should indicate the survey version on the Take Delegation Survey page so that, say, if we change it mid-stream, we can ask OSU to verify that they are seeing the survey version we think they should be seeing.

kaitlyn-sharo (approved):
Looks good to me, aside from the console.logs which darren already mentioned. I also like the idea of putting the survey version on the front page of the survey so OSU can confirm it is correct, especially since there are sometimes caching issues. 

Some other things I noticed that might not necessarily related to this PR (that's up to you) but might need tickets:
- Version 1 in the comparison page does not show both medics with buttons like the other versions do
- Version 1, there's another section that is clearly supposed to have a button for review, but does not show it
- Version 1 does not have a progress bar
- Version 0 has lost its images, and the "Supplies" header has things that are not supplies
- Version 0 does not have options to review medic decisions, same as version 1
- Omnibus is also broken for version 0
- I'm also missing images in version 2, wondering if this is residual from when I ran Phi's ingest script that took away all my MRE images, but I thought I reset my database since then...
- Version 2 also missing medic review buttons
- Non-supplies also in supplies section for Version 2 (probably because the header used to be supplies/constraints)
- Omnibus version 2 broken
- Same for Version 2.1 as Version 2, aside from omnibus since 2.1 doesn't have omnibus
- Version 3 images missing, thinking this is probably on my end, but maybe someone else can confirm?
- Review buttons also missing in version 3
- Version 4 still looks normal
I only mention these old version problems because if the point of this is to be able to switch to an earlier version quickly, we don't want it to be broken. I'm happy to take a look into these if you want, just lmk

kaitlyn-sharo:
I reset my database with the latest from the aws backups and all my images are still missing, aside from Version 4

brianpippin:
I just want to add here that we aren't as worried about older versions than 3 working at this point and that we would only want to be able to toggle some versions if we are running multiple experiments, but I think working out some of the basic issues with the images and what not will help with that.

dereknop:
@kaitlyn-sharo I believe I fixed everything you mentioned. The issue with the pictures was a real headache. I think something got broken with the way earlier survey versions patient imgUrl was mapping to the entries in `delegationMedia`. I winded up needing to re-run the one time setup script in ingest (versions 0-2.1) using `update_survey_config.py` and the `survey_setup_script.py` for version 3. I also added that progress bar to version one so I'll open up an ingest pr. @nextcen-dgemoets I added an indicator in the bottom right hand of the survey page that says the current version (if this looks to distracting to participants I'll edit it). The omnibus and review modals on the comparison pages for 0.0-3.0 should now be functional. 

I'll write explicit instructions in ingest pr for what I ran in ingest to get the images properly loading in

phile-caci (approved):
LGTM, changed it to version 3 and the survey looked like MRE stuff, also the download from survey results defaulted to version 3, and same for setting it to version 4

nextcen-dgemoets:
Looks really good, Derek.  I didn't test everything in Kaitlyn's list, but the survey indicator worked and the progress bars are nice in the old surveys.  I didn't notice the survey version at first, but I think that's the idea-- not to be intrusive.  The only thing that caught me off guard a little was that the progress indicator in the legacy surveys had dashes between each circle, but in 4.0, they don't.  But I think this is because in 4.0 the number is variable.  It was the same on prod, so it must be fine.

Re-approved!

brianpippin (approved):
Everything looking good here!

PR #127, opened 2024-09-30
Participant-Level variable update

kaitlyn-sharo:
Added the DRE version of the variable definition table for the participant-level data (http://localhost:3000/humanSimParticipant). MRE and DRE should have different variable tables and downloads, matching the variables shown in the header on each respective page.

Also added the delegation column values, which for DRE is calculated by the number of times a participant chose aligned in the comparison page over baseline/misaligned. For those Parallax runs with only two adms, if the non-baseline adm is misaligned, the question is ignored, as we have no good comparison.


kaitlyn-sharo:
> I got a crash. Do we need more optional chaining for my broken DB?
> 
> ```
> Cannot read properties of undefined (reading 'find')
> TypeError: Cannot read properties of undefined (reading 'find')
>     at populateDataSet (http://localhost:3000/static/js/bundle.js:3043:85)
>     at http://localhost:3000/static/js/bundle.js:3581:83
>     at commitHookEffectListMount (http://localhost:3000/static/js/bundle.js:328589:30)
>     at commitPassiveMountOnFiber (http://localhost:3000/static/js/bundle.js:330087:15)
>     at commitPassiveMountEffects_complete (http://localhost:3000/static/js/bundle.js:330054:13)
>     at commitPassiveMountEffects_begin (http://localhost:3000/static/js/bundle.js:330044:11)
>     at commitPassiveMountEffects (http://localhost:3000/static/js/bundle.js:330034:7)
>     at flushPassiveEffectsImpl (http://localhost:3000/static/js/bundle.js:331917:7)
>     at flushPassiveEffects (http://localhost:3000/static/js/bundle.js:331870:18)
>     at http://localhost:3000/static/js/bundle.js:331685:13
> ```
> 
> ```
> The above error occurred in the <AggregateResults> component:
>     at AggregateResults (http://localhost:3000/static/js/bundle.js:3537:3)
>     at div
>     at HomePage (http://localhost:3000/static/js/bundle.js:9622:5)
>     at Home (http://localhost:3000/static/js/bundle.js:6894:3)
>     at Route (http://localhost:3000/static/js/bundle.js:340426:29)
>     at Switch (http://localhost:3000/static/js/bundle.js:340595:29)
>     at div
>     at div
>     at Query (http://localhost:3000/static/js/bundle.js:22921:24)
>     at Router (http://localhost:3000/static/js/bundle.js:340098:30)
>     at App (http://localhost:3000/static/js/bundle.js:7086:5)
>     at ApolloProvider (http://localhost:3000/static/js/bundle.js:22805:19)
>     at Provider (http://localhost:3000/static/js/bundle.js:616677:3)
> ```

This doesn't have a line number in it, but I didn't add any "find" operations in this PR. Did you rerun the startup script? I didn't mention it, because the server isn't changed in this PR, but it is changed in the base PR, so maybe you need to do that. Give it a try and let me know. If it doesn't work, please give me the line number from the console.

nextcen-dgemoets:
Sorry, I copied one stracktrace from the dev console, but the other from the screen.  Here's the other proper stack trace:

nextcen-dgemoets:
Now that I look at it more closely, I think I got a crash on that line before in another ticket (DataFunctions.js:764).

kaitlyn-sharo:
> Now that I look at it more closely, I think I got a crash on that line before in another ticket (DataFunctions.js:764).
> 
> [EDIT: and yes, I always re-rerun the startup script.]

Okay, if it's the same line, then I just need to merge that one into this

kaitlyn-sharo:
> Everything looked good except I had one entry with a PID of 'krc'. I am not sure which query this is pulling from since the query is three combined
> 
> ```
> const GET_SURVEY_RESULTS = gql`
>     query GetAllResults($evalNumber: Float!){
>         getAllSurveyResultsByEval(evalNumber: $evalNumber),
>         getAllScenarioResultsByEval(evalNumber: $evalNumber),
>         getAllSimAlignmentByEval(evalNumber: $evalNumber)
>     }`;
> ```
> 
> Maybe we should pull down the participant log collection and filter out any entries with a pid not in the log.
> 
> ```
> const GET_PARTICIPANT_LOG = gql`
>     query GetParticipantLog {
>         getParticipantLog
>     }
> `
> ```
> 
> Inside of `AggregateResults`: `const { loading: participantLogLoading, error: participantLogError, data: participantLogData } = useQuery(GET_PARTICIPANT_LOG);`
> 
> And then the second useEffect:
> 
> ```
> React.useEffect(() => {
>         if (data?.getAllSurveyResultsByEval && data?.getAllScenarioResultsByEval && participantLogData?.getParticipantLog) {
>             const fullData = populateDataSet(data);
>             const participantIds = participantLogData.getParticipantLog.map(entry => String(entry.ParticipantID));
>             const filtered = fullData.filter(entry => participantIds.includes(entry.ParticipantID));
>             
>             setFullData(filtered);
>             setAggregateData(getAggregatedData(filtered));
>             const xtraData = getChartData(filtered);
>             setChartData(xtraData);
>             setKdmaScatter(xtraData.scatter);
>         }
>     }, [data, participantLogData, error, loading, participantLogLoading, participantLogError]);
> ```
> 
> This worked for me. I know this isn't part of this ticket. Just mentioning so we ensure the table doesn't contain noise

Weird...that's my pid that I typically use but I'm not seeing it on my end, so I thought it was filtering everything out...I'll look into it

dereknop:
> Weird...that's my pid that I typically use 

I believe my most recent backup is a copy of yours, so that checks out

nextcen-dgemoets:
Yeah I've got junk in my DB also which is giving me some results I'm not expecting.  e.g., my `dre_participant_data.xlsx` has a bunch of spurious fields like:
- Sim1,	Sim2,	SimOrder,	TextSimDiff,	ST_KDMA_Text,	ST_KDMA_Sim	, ST_AttribGrp_Text	, ST_AttribGrp_Sim	, AD_KDMA_Text	, AD_KDMA_Sim	, AD_AttribGrp_Text	, AD_AttribGrp_Sim

Maybe a filter would help.  These are only in weird rows with no `Date`, for example.

kaitlyn-sharo:
> Yeah I've got junk in my DB also which is giving me some results I'm not expecting. e.g., my `dre_participant_data.xlsx` has a bunch of spurious fields like:
> 
> * Sim1,	Sim2,	SimOrder,	TextSimDiff,	ST_KDMA_Text,	ST_KDMA_Sim	, ST_AttribGrp_Text	, ST_AttribGrp_Sim	, AD_KDMA_Text	, AD_KDMA_Sim	, AD_AttribGrp_Text	, AD_AttribGrp_Sim
> 
> Maybe a filter would help. These are only in weird rows with no `Date`, for example.

So weird, I don't know why mine isn't showing this. Looking into it still

kaitlyn-sharo:
@dereknop @nextcen-dgemoets This should be fixed now. After taking the survey with an invalid pid and then refreshing the page, I _finally_ was able to reproduce the behavior. I changed it where the data is created rather than where the data is displayed to avoid unnecessary processing

nextcen-dgemoets:
It did clear out one of my spurious PIDs, but I still have entries from PID 202409101 and 202409102 that have those extra fields listed above (they show up in the DL but not in the dashboard table).  I don't know where that data comes from.  Maybe filter out if the Date is empty?

Also, I thought we sorted by ParticipantID, but maybe that was another table.  Not strictly required, of course.

nextcen-dgemoets:
Looks a lot better after the latest change.  There's still a spurious row with PID `202409103` that has a date but is mostly empty.  But it doesn't have all of those extra columns listed above, so the output looks *almost* clean.  That raw was created last Thursday 9/26, so it was probably from some testing of mine.

nextcen-dgemoets (approved):

PR #132, opened 2024-10-02
Itm 617 exploratory analysis

kaitlyn-sharo:
Added new Exploratory Analysis section with variable downloads and table skeletons. Tables will be filled as ingest scripts are created to calculate missing scores (or in the case of table 8, analyzing open world)

The 616 ingest script should be run first. Rerun the startup script as well.


phile-caci (approved):
Everything looks like it's working, tested all the filters, download, download definition.

PR #134, opened 2024-10-03
started adding open world data to RQ8

kaitlyn-sharo:
Adds the correct rows to the RQ8 table. SInce no analysis has been done on the open world data, most columns are still empty, but at least now the table has some data.

https://github.com/NextCenturyCorporation/itm-ingest/pull/52
Run this ingest script first!

Make sure that the other sim-related pages (play-by-play and probe values) are unaffected by the addition of open world data to the database


phile-caci (approved):
LGTM!

image

Also spot checked the play-by-play and probe values between prod and my local, didn't see any differences

nextcen-dgemoets (approved):
New tables, downloads, and variable definitions looked good, and didn't encounter any issues with other tables and data.

PR #135, opened 2024-10-03
Itm 615

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-615)

I believe I hit all of the UI updates in this ticket. Please let me know if anything stands out as incorrect or clunky.


nextcen-dgemoets:
- Regarding "remove the word 'Scenario' from the beginning of every scenario", I'm going to guess that @brianpippin meant only in Survey 4.x.  It gives odd-looking results for Survey's 1.x - 3.x.
- I have a seemingly spurious 0.x in the Survey Version of the Delegation Survey Results page-- it not on prod.  Unless this has something to do with changing the survey version on the admin page?  I currently have it set to 4, but maybe this affected the survey results page.  Something to look for.
- I also found in the Tabulated Survey Results table that I have a Survey Version `1.0` and `3.0` (not to be confused with `1` and `3`) with 1-2 spurious rows (e.g., PID `fghj`).  These are not in prod, so I guess it's some extra garbage in my local DB.
- Text Scenario Results looks good, except that I noticed a spurious `vol-dre-1-eval` scenario with the MRE data (PID of "ewfsdfsd").  This is on prod (and local).
  - Minor nit:  in the new instructions, suggest changing last sentence to "In the Participants view, a download button is available to export the data."
- I really like the changes in Human Sim Probe Values.  But it looks like it broke the MRE table headers.
  - I've never understood referring to Adept Scenario 1, 2, and 3 to mean MJ2, MJ4, and MJ5, but this seems to be an accepted thing.
  - I guess if we *are* going to refer to MJ2 as "Scenario 1", maybe its tables should come first-- current order is 2, 3, 1.  And maybe the tables with QOL-VOL-1 should come before those with QOL-VOL-2 and QOL-VOL-3.  But that's pretty minor.
- Re: Play by Play Humans in the Sim, this is also a nice change!  There appears to be an extra space after the header in the MRE version.

phile-caci:
Looks like most of Brian's/Darren's comments have been address, This is the only thing I noticed:
image

dereknop:
Merge conflicts resolved, survey results page charts should be updated properly when changing scenarios, and humanProbeData page should be working for DRE and MRE.

dereknop:
@nextcen-dgemoets That entry you noticed that was on prod lumped in with metrics eval data for a dre scenario was pruned out with my last ingest pr for removing duplicate text based scenario entries.

kaitlyn-sharo (commented):
image


This is just a personal preference, but I think this new header would look better aligned with the table.

One other thing, the changes made to aggregateResults should really go in DataFunctions. You can see lines 683-687 is where the participants were separated into adept groups. This file is for data manipulation/processing, whereas aggregateResults is more the UI-side. If it's not too hard, I think it's worth moving it over.

All the changes look great though, and everything in the ticket was hit.

dereknop:
> One other thing, the changes made to aggregateResults should really go in DataFunctions. You can see lines 683-687 is where the participants were separated into adept groups. This file is for data manipulation/processing, whereas aggregateResults is more the UI-side. If it's not too hard, I think it's worth moving it over.


Fair point, I'll move those functions to the DataFunctions file and align that header

kaitlyn-sharo (approved):
It looks like there's still some data manipulation in aggregateResults, but if it becomes a problem we can move it later. Everything looks good.

PR #136, opened 2024-10-03
Dev to Main

brianpippin:
Started dev to main because the remaining tickets could be deployed tomorrow or Monday if necessary.


phile-caci (approved):
test Dev UI with all the data ingest, clicked through most things, looks good to me

nextcen-dgemoets (approved):
Can't say I remember all of the changes since last time (though I did scan the commits), but I didn't see any problems clicking and scrolling around.

kaitlyn-sharo (commented):
I'm having to refresh some of the RQ pages to get data to show up. Is anyone else seeing this? Just sometimes it says "showing 0 of x entries based on filters" but no filters have been chosen, then when I refresh it works. I hadn't seen this before

nextcen-dgemoets:
> I'm having to refresh some of the RQ pages to get data to show up. Is anyone else seeing this? Just sometimes it says "showing 0 of x entries based on filters" but no filters have been chosen, then when I refresh it works. I hadn't seen this before

I didn't see that, at least on the pages I viewed.

kaitlyn-sharo:
> I'm having to refresh some of the RQ pages to get data to show up. Is anyone else seeing this? Just sometimes it says "showing 0 of x entries based on filters" but no filters have been chosen, then when I refresh it works. I hadn't seen this before

I was experiencing this consistently and determined it was because I was missing formattedData in the useEffects dependency array. Probably things were loading in a different order than I expected, whether due to a slow computer or whatever. This has fixed the problem for me.

kaitlyn-sharo (approved):
I think all the data I added is showing up now and the survey change derek made that's in this works for me too

PR #138, opened 2024-10-07
Itm 623 data fix

kaitlyn-sharo:
The Exploratory Analysis docs had a typo that said to base everything off of the Delegation Survey. Instead, it was supposed to be based off the Text-based responses. This PR fixes the tables so that they are based off of the Text-based responses. RQ4 is unchanged, but RQ5, 6, and 8 have been. 

Run https://github.com/NextCenturyCorporation/itm-ingest/pull/55 first in order to get the correct most/least aligned comparison values for the tables. 

There are still some empty columns, like all the Group Target columns in RQ5, the last column in RQ6, and most columns in RQ8. Those changes are coming!


nextcen-dgemoets (approved):
The values have changed from prod.  I have no real way to know that the changes are good/correct, just as I didn't know that the values were wrong/bad before, but the database was updated and it's reflected in the dashboard.

phile-caci (approved):
After running: https://github.com/NextCenturyCorporation/itm-ingest/pull/58
Ran: [itm-628-group-targets, ](https://github.com/NextCenturyCorporation/itm-ingest/pull/57)
which includes this branch 

Looks good!

PR #139, opened 2024-10-07
Itm 619 sim text align

kaitlyn-sharo:
Displays the results of https://github.com/NextCenturyCorporation/itm-ingest/compare/itm-619-compare-sim-vs-text?expand=1 in table RQ6

Currently, IO and MJ are the same number. Once Adept's server is fixed, these numbers will be separated. There is already a ticket for this.


nextcen-dgemoets:
I'm not seeing any values in the "Alignment score (Participant_Text|Participant_Sim)" column.  Is this because of the timestamp errors I was getting in the 619 ingest PR?

nextcen-dgemoets (approved):
The data was populated in the table and in the download.

brianpippin (approved):

PR #140, opened 2024-10-07
Itm 626

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-626)

There was an issue with duplicate documents being uploaded for the text-based scenarios. The duplication would happen when one participant completed their text-based scenarios, followed by someone resetting the page by pressing 'M'. If another participant then completed their text-based scenarios, the previous participant's Adept documents would mistakenly be uploaded again (in addition to the current participant's documents). 

This is a little bit of a headache to test. The easiest way I'd suggest is to take a PID that isn't in your db already i.e `20249241` and click through the scenarios (sorry). 3 new adept documents should be added to `userScenarioResults`, 2 new soartech ones should be added. 

Then take another PID from the log (faster if you pick one with Adept as `text-1` i.e `20249243`). All you have to do this time is click until you are through the three adept scenarios. At which point, you can check your Mongo db to see that 3 new adept documents have been added. 

If you repeat these steps on dev, you should experience the bug. That is, the duplicate documents being uploaded.

### DIFF
--- a/dashboard-ui/src/components/TextBasedScenarios/TextBasedScenariosPage.jsx
+++ b/dashboard-ui/src/components/TextBasedScenarios/TextBasedScenariosPage.jsx
@@ -179,7 +179,11 @@ class TextBasedScenariosPage extends Component {
         const { scenarios, currentScenarioIndex } = this.state;
         if (currentScenarioIndex < scenarios.length) {
             const currentScenario = scenarios[currentScenarioIndex];
-            this.loadSurveyConfig([currentScenario], currentScenario.title);
+            const newStartTime = new Date().toString();
+            
+            this.setState({ startTime: newStartTime }, () => {
+                this.loadSurveyConfig([currentScenario], currentScenario.title);
+            });
         } else {
             this.handleAllScenariosCompleted();
         }
@@ -569,13 +573,7 @@ class TextBasedScenariosPage extends Component {
     onAfterRenderPage = (sender, options) => {
         const pageName = options.page.name;
         const currentTime = new Date();
-
-        if (!this.state.startTime) {
-            this.setState({
-                startTime: currentTime.toString()
-            });
-        }
-
+    
         if (this.survey.currentPageNo > 0) {
             const previousPageName = this.survey.pages[this.survey.currentPageNo - 1].name;
             const startTime = this.pageStartTimes[previousPageName];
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
I followed the instructions on both `development` and the feature branch.  On development, duplicate records were created for 20249241's ADEPT scenarios, and on the feature branch, no additional records were created!

dereknop:
> I noticed that it doesn't save the last few scenarios unless I navigate away from (i.e., I clicked the `Home` menu button) 

Thanks for this catch. It uploaded for me without navigating away but it took a little bit (maybe 15-20 seconds). I think maybe I should include a loading spinner and a message that warns the user not to close the browser while the results are processed and uploaded. Open to any other ideas though

nextcen-dgemoets:
Yeah I noticed that some scenarios were populated after a delay.  A warning and/or spinner sounds like a good idea.

phile-caci:
Hi @dereknop , I see the Uploading screen, but after it completes, I'm only seeing 4 documents now?  It should still be 5? (I checked the db for several minutes into the Thanks You screen )

dashboard> db.userScenarioResults.count({"participantID": "20249241"})
4

kaitlyn-sharo (commented):
It worked in that I did not get duplicate entries (while I did get duplicates on dev), but every time I tried on this branch, IO1 did not appear in my database. I waited to make sure it wasn't just a lagging server, but nothing ever came in. It was always IO1. Ran with the exact ids mentioned in the description

dereknop:
Ugh... looks like one of those last commits must have had a bug. Looking into it

phile-caci (approved):
Thanks Derek, I think we're looking good now, everything saves as expected, and no duplicates. (Just a note, saving documents 4 and 5 takes a while, close to a minute and a half on my setup)

dereknop:
> (Just a note, saving documents 4 and 5 takes a while, close to a minute and a half on my setup)

This is fair, and I noticed it as well. I am not sure how to speed up the requests and responses to the ST server. The silver lining is that when a participant is going through they will take longer than us just quickly clicking through. So that 4th document will most likely be done uploading before they finish the 5th scenario. That would hopefully make that final wait time well under a minute.

PR #141, opened 2024-10-09
Itm 628 group targets

kaitlyn-sharo:
Run https://github.com/NextCenturyCorporation/itm-ingest/pull/57 first. This will add group target data to the database.

- Table 2.1 should now have lots of fun data to look at.
- Did some cleanup with a utils file that stores reused components
- Table 2.2-2.3 now has group target rows populated
- RQ5 has missing columns about group targets populated

This will require a rerun of the startup script


phile-caci (approved):
After running: [ITM-629_RePopAdeptSession](https://github.com/NextCenturyCorporation/itm-ingest/pull/58)

Ran this without issue

brianpippin (approved):
Kaitlyn verified we are getting same results.

PR #146, opened 2024-10-21
Prevent x.0 in survey version recorded from survey results to prevent issues on data presentation pages

dereknop:
After changing to survey version 3 on the admin page, future survey results were populating on the survey results page as `3.0` instead of `3`, creating two separate filters for what should be grouped together.

Test this out on dev, you will only need to click through a couple pages of the survey to see this behavior on the results page. Next, switch to the feature branch, delete the test document from the `surveyResults` collection in your database, and repeat the process. Verify that new results now show as version "3" (not "3.0") and that there's only one filter for version 3.

### DIFF
--- a/dashboard-ui/src/components/Survey/survey.jsx
+++ b/dashboard-ui/src/components/Survey/survey.jsx
@@ -11,7 +11,7 @@ import { OmnibusComparison } from "./omnibusComparison";
 import gql from "graphql-tag";
 import { Mutation } from '@apollo/react-components';
 import { useQuery } from 'react-apollo'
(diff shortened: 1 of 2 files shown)


nextcen-dgemoets (approved):
Works exactly as described-- In `development` it created the `3.0` version (in the table), but in the feature branch, it stayed at version `3`.

phile-caci (approved):
LGTM,
Tested Dev branch, ran a del survey and the del results had a 3.0 and a 3 filter in the  `View Tabulated Data`
Tested this branch, ran a del survey, and the del results only had a 3 filter in the `View Tabulated Data`

PR #150, opened 2024-10-30
trial ids based on order log

kaitlyn-sharo:
RQ 1/3/4 trial ids used to be counted simply based on loading order, but we added some randomization within each set (with a set being baseline, aligned, misaligned). Trial IDs should be based off of that randomization instead. 

This grabs the order log from the delegation survey (if it exists, since it does not exist for the first participants) and checks the page order to determine trial id.


nextcen-dgemoets (approved):
Comparing the downloaded RQ1/3 data from `development` to this branch, I see subtle differences in `Trial_ID` in particular.  So as far as I can tell, that means it works!

brianpippin (approved):
LGTM!

PR #152, opened 2024-10-31
added eval toggles to rq pages

kaitlyn-sharo:
Not sure if we are going to merge this yet, but it can at least be approved and sit until we have real Phase 1 data. 

Added a select dropdown to all the RQ pages for DRE or Phase 1. If you select Phase 1, nothing should appear in the tables. If you want to take an extra step in testing, go into compass and manually change a few eval numbers from 4 to 5 and see the tables in Phase 1 begin to populate.
- Change an entry in userScenarioResults (evalNumber is at the top level)
- Change an entry in surveyResults (evalNumber is in "results")
- Change an entry in test (any entry)
- Change an entry at the end of test, making sure it is a group target (check the last item in history to see that it is one of the group targets)

Not all the data will be populated, but if you change these, each table should show at least 1 entry.


phile-caci (approved):
LGTM!

Tested without Eval5 data, no issues

Switched data to Eval5, no eval 4 data, no issues with empty Eval4, RQ1-3 tables were populated
```
db.userScenarioResults.updateMany({"evalNumber": 4}, {$set: {evalNumber: 5}})
db.surveyResults.updateMany({"results.evalNumber":4},{$set: {"results.evalNumber": 5}})
db.test.updateMany({"evalNumber": 4}, {$set: {evalNumber: 5}})
```

dereknop (approved):
Looks good. One small thing I noticed is that if I toggle to Phase 1 on RQ1 and then go to RQ2 it defaults back to DRE. I would think if someone wants to see phase 1 data they would want to continue viewing it on the next page. This is pretty insignificant so it may not be worth effort to track this across the different rq pages. I do see that the default value is `ALLOWED_VALUES[0]`. This makes me think maybe we should have the current eval set in the `.env` or somewhere in the store and just pull from there so the code doesn't need to be changed when the eval changes.

kaitlyn-sharo:
> Looks good. One small thing I noticed is that if I toggle to Phase 1 on RQ1 and then go to RQ2 it defaults back to DRE. I would think if someone wants to see phase 1 data they would want to continue viewing it on the next page. This is pretty insignificant so it may not be worth effort to track this across the different rq pages. I do see that the default value is `ALLOWED_VALUES[0]`. This makes me think maybe we should have the current eval set in the `.env` or somewhere in the store and just pull from there so the code doesn't need to be changed when the eval changes.

It's true the default value is ALLOWED_VALUES[0], but that should never be used because selectedEval should be used always. I think that's just a fallback, really. I can look into putting this in the store, but I imagine it'll be used like all the other pages, where once we have data, we'll just set the selected eval to 5 and that'll be the default that everyone will look at

dereknop:
> we'll just set the selected eval to 5 and that'll be the default that everyone will look at

Gotcha, makes sense 👍

brianpippin:
We should get this up to dev as well to make sure we are ready for Eval 5 data.

PR #154, opened 2024-11-01
logic to add new pids when log runs out

kaitlyn-sharo:
We want to be able to dynamically create new entries in the participant log when the current log runs out. To test this, start by setting all the entries in participantLog to have claimed: true. Then enter some emails in this page http://localhost:3000/participantText

See that new entries are populated following the pattern for Mil or Civ, listed at the top of index.js (there are 12 ordering options for each). 

Open two or more pages at the same time and enter details on all of them. Then, one-by-one press "Start" and ensure that no duplicate ids occur. 

Also, revert your database to what it was before you turned claimed: true on for everyone and make sure that the old way continues to work without race conditions as well (no two participants should be able to get the same pid - EVER). Also check that the PID lookup page still works

### DIFF
--- a/dashboard-ui/src/components/App/index.jsx
+++ b/dashboard-ui/src/components/App/index.jsx
@@ -47,12 +47,16 @@ import { PidLookup } from '../Account/pidLookup';
 
 const history = createBrowserHistory();
 
-const GET_SURVEY_VERSION_AND_PARTICIPANTS = gql`
+const GET_SURVEY_VERSION = gql`
   query GetSurveyVersion {
     getCurrentSurveyVersion
-    getParticipantLog
   }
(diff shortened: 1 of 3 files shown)


phile-caci (approved):
LGTM, 

Updated all the participantLogs.claimed to true
`db.participantLog.updateMany({}, {$set: {claimed: true}})`

Logged in at /participantText with new email, completed text scenario

Logged in as Admin

PI Lookup:
`PID: 20249261` <- New PID

kaitlyn-sharo:
Replaced with https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/157/files - same exact code!

PR #159, opened 2024-11-11
Itm 670

dereknop:
Depends on running the [ingest pr](https://github.com/NextCenturyCorporation/itm-ingest/pull/70) first. 

http://localhost:3000/review-text-based

You should see the 5 adept scenarios under Phase 1 and 6 Soartech scenarios under Phase 1. 

Click through and make sure that there are no obvious errors. You can check using the yaml files located in ingest `text_based_scenarios/phase1_yaml_files`

One thing to look out for is the behavior on MJ5. Make sure the patient you assess first has normal breathing and heart rate vitals and the other patient has fast breathing and heart rate vitals. These vitals should remain until after the probe that asks you about how much gauze each patient receives. 

All of the scenarios should have images besides the two Adept training scenarios (OI1, MJ1)

Note: these labels for each probe are for debugging purposes to make it easy to cross-reference what you are looking at with the YAML files. This will be removed before merging (or after TA1 reviews)
Screenshot 2024-11-11 at 4 04 24 PM

### DIFF
--- a/dashboard-ui/src/components/App/setupUtils.js
+++ b/dashboard-ui/src/components/App/setupUtils.js
@@ -49,7 +49,13 @@ export function setupTextBasedConfig(data) {
                 for (const el of page.elements) {
                     if (Object.keys(el).includes("patients")) {
                         for (const patient of el.patients) {
-                            const foundImg = data.getAllTextBasedImages.find((x) => (x.casualtyId?.toLowerCase() === patient.id?.toLowerCase() && x.scenarioId === page.scenario_id));
+                            /*
+                            soartech has different images for same casualty id's depending on scenario so we need to make sure casualty id matches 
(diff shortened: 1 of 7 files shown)


phile-caci (approved):
LGTM!

Didn't see any obvious errors
MJ5 only has normal for 1st assessed, vitals changed after gauze
All scenarios have images except oi1/MJ1

nextcen-dgemoets (approved):
I didn't see any issues.  MJ5 vitals looked right based on description in PR.

I noticed that we're using the TA3 filename convention for the ADEPT scenarios.  I guess we had to do *something* considering they would be indistinguishable from DRE scenarios.  Maybe we should have asked both TA1s to update their filenames and scenario IDs for the final eval.

Also just a nit on this page, but Adept instead of ADEPT.

PR #162, opened 2024-11-15
add option to download filtered data only (rq pages)

kaitlyn-sharo:
Check out the RQ tables. The page should look the same until you turn on a filter. When a filter turns on, the "Download Filtered Data" button should show up. I've put it under the Download All button to help reduce the length-wise clutter in the filter section. When filters are off, the button should disappear again. 

Make sure all the table buttons work as expected.

### DIFF
--- a/dashboard-ui/src/components/DRE-Research/tables/download-buttons.jsx
+++ b/dashboard-ui/src/components/DRE-Research/tables/download-buttons.jsx
@@ -1,0 +1,17 @@
+import React from "react";
+import { exportToExcel } from "../utils";
+import '../../SurveyResults/resultsTable.css';
+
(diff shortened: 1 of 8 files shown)


kaitlyn-sharo:
> Looks great-- love this feature. I wanted to request it when the filters were first added, but I try not to be any more annoying than I already am.
> 
> BTW, unrelated to this ticket, I noticed that the downloaded filename for RQ4 was `RQ-1_and_RQ-3 data.xlsx` instead of `RQ-4 data.xlsx`.

They are the same graph. It's technically 1, 3, and 4, but that's a lot for a name and 4 is part of exploratory analysis, so I think it's okay to keep it as 1 and 3

PR #164, opened 2024-11-19
Itm 698

dereknop:
New ADEPT comparison before forced choice questions and small logic change for participant log change. 

If you click through the survey to an adept comparison page (use pid 'test' and click a few pages through), you should see something like: 
Screenshot 2024-11-19 at 11 04 30 AM

Clicking on it should have the medic decisions of the two medics in the forced-choice question below :

Screenshot 2024-11-19 at 10 56 31 AM

This change is only for adept, soartech remains the same as DRE.


dashboard-ui/src/components/App/index.jsx
```diff
@@ -296,7 +296,13 @@ export class App extends React.Component {
                     });
                 }
                 else {
-                    const newPid = Math.max(...pLog.filter((x) => x.Type === classification && !["202409113A", "202409113B"].includes(x['ParticipantID'])).map((x) => Number(x['ParticipantID']))) + 1;
+                    // still want to record distinction between civ and mil but it should no longer effect the actual pid
+                    const newPid = Math.max(...pLog.filter((x) => 
+                        (classification === 'Mil' || classification === 'Civ' ? 
+                            (x.Type === 'Mil' || x.Type === 'Civ') : 
+                            x.Type === classification) &&
```
kaitlyn-sharo, line 303: I'm not sure any of this classification stuff is needed. I'd remove it for now and then later when we add the online classification we can update this, but I think we just want to filter out those two strings and that's it

kaitlyn-sharo (commented):
This looks great! To be clear before the rest of my comments, the only change I'm looking for is the one comment I left on the code.

I feel like everything got super big all of a sudden, but I don't see that in the css changes, so it's probably just me again. I feel like dashboard sizes are always changing. I have to zoom out to 70-80% to make it comfortable to use.

The new comparison is awesome. I wonder if in the future we could make them scroll together so you don't have to go back and forth scrolling each one to get the action "timing" to match up. But that's just a personal preference

dashboard-ui/src/components/App/index.jsx
dereknop: True. I'll leave in the filter for ["202409113A", "202409113B"] because Mongo says those are the max numbers but remove the rest of the filter

kaitlyn-sharo (approved):

dereknop:
> I wouldn't want them to use the actual browser back button.

Good point, I just checked to make sure they still get the warning if they hit the browser button.

PR #167, opened 2024-11-21
ADM Target Setup (ph1)

kaitlyn-sharo:
Cleaned up survey.jsx by moving around some variables/functions to other files.
Also set up adms for phase 1. Check the targets against https://nextcentury.atlassian.net/wiki/spaces/ITMC/pages/3476946948/Phase+I+Evaluation+Delegation+Materials

You'll need to change your survey version to 5 using the admin page before trying this out.

Run through with the default adms and a few pids to test that everything looks good in terms of scenarios/adms/targets.

### DIFF
--- a/dashboard-ui/src/components/DRE-Research/utils.js
+++ b/dashboard-ui/src/components/DRE-Research/utils.js
@@ -1,7 +1,7 @@
 import * as FileSaver from 'file-saver';
 import XLSX from 'sheetjs-style';
-import { admOrderMapping, delEnvMapping } from "../Survey/survey";
 import { isDefined } from "../AggregateResults/DataFunctions";
+import { admOrderMapping, getDelEnvMapping } from '../Survey/delegationMappings';
 
 
 export const ADM_NAME_MAP = {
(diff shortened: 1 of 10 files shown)


dereknop:
fixing issue with soar tech comparison before forced chocie

dereknop (approved):

phile-caci:
Checking if I did some thing wrong, I got this in the meantime whiel going through a survey:
image

dereknop:
> Checking if I did some thing wrong, I got this in the meantime whiel going through a survey: image

Darren ran into the same thing earlier. I had him delete survey config 5 from the `delegationConfig` and delete his `admMedics` collection and re run the ingest steps and it worked itself out.

nextcen-dgemoets:
In the category of "we have higher-priority items than this currently," I noticed that when I collapse the scenes in the delegation survey, then click on a checkbox at the bottom, all of the scenes expand again.

nextcen-dgemoets (approved):
The alignment targets looked right.
The baseline mappings match the Confluence page.
I ran through the delegation survey without any errors.

So I guess it LGTM!

phile-caci:
@dereknop , I'm seeing the same issue, just with ADEPT after:

db.delegationConfig.deleteMany({"survey.version":5})
db.admMedics.deleteMany({"evalNumber":5})

and rerunning ingest.  Is there something else I should be clearing?

phile-caci (approved):
Everthing looks good to me after Derek's (thanks Derek) latest update!

dereknop:
> We definitely need clean steps so the prod deployment goes smoothly and maybe if this is what needs to be done, we do this first before the deployment.

Phi and I worked out all of the steps for prod this morning

PR #169, opened 2024-11-22
Text Scenario Test Page

kaitlyn-sharo:
NOTE: PROBABLY NOT FOR THIS ROUND, so not a high priority review (but possibly helpful in the future!)

This PR creates a way for us and OSU to run through the text scenarios without polluting data. The /participantText and /login pages are only available when logged out, currently. This makes a third page, /participantTextTester, that is only available when logged in as an experimenter (OSU) or as an admin (us). To access this page, look at your user dropdown (right under PID Lookup). 

Clicking on it should open up the login page while still keeping your header. 

~~You will see the radio option has changed so ONLY tester is available. Tester has all 24 variations of scenario-ordering (whereas mil and civ only have 12 each).~~

**Update: Now that we do not have classification on the email participant page, this will automatically set entries through this special route to "Test".** 

Check that in the participant id log, when the pid is generated, the type shows up as "Test" in mongo.

Make sure everything else works smoothly.

### DIFF
--- a/dashboard-ui/src/components/Account/login.jsx
+++ b/dashboard-ui/src/components/Account/login.jsx
@@ -205,107 +205,111 @@ class LoginApp extends React.Component {
                 <div className="row justify-content-center align-items-center h-100 center-container">
                     <div className="login-container">
                         <Tabs className='p-1' defaultActiveKey={this.props.isParticipant ? 2 : 1}>
-                            <Tab eventKey={0} title="Create Account">
-                                <div>
-                                    <div className="sign-in-instructions">
-                                        <h4 className="sign-in-header">Create Account</h4>
-                                        <p className='subtext'>Create an account to view results.</p>
-                                    </div>
-                                    <form onSubmit={this.createAccount}>
-                                        <div className="form-group">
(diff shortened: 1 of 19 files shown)


dereknop (approved):
Worked great for me. I see the type marked as tester in the participant log. One thing I noticed (but not part of this ticket) is using an existing email generates the new pid in the participant log even if you hit 'cancel'. So on the results pages I think we are already checking if the results match to a participant id in the participant log, I suppose we will have to add to that condition to specify that the entry it maps to in the participant log cant be of type 'tester'

kaitlyn-sharo:
> Worked great for me. I see the type marked as tester in the participant log. One thing I noticed (but not part of this ticket) is using an existing email generates the new pid in the participant log even if you hit 'cancel'. So on the results pages I think we are already checking if the results match to a participant id in the participant log, I suppose we will have to add to that condition to specify that the entry it maps to in the participant log cant be of type 'Test'

Yes, I agree. I had thought of that but must have forgotten to implement it. I will add a ticket.

phile-caci (approved):
LGTM!

brianpippin (approved):
Saw no issues with this one.

PR #171, opened 2024-11-22
Overlap with baseline adept

dereknop:


dereknop:
This should have logs that tell you the exact adms being loaded in for the 4 sets. This fixes an issue where if the most aligned adm was overlapping with the subset of baseline ones it wasn't assigning a different target as the aligned adm.

dereknop:
An example to test this would be my test run. If you backed up your db with what phi sent, you can see that the survey result for pid `202411301` does not properly load in a new aligned adm (0.3 and 0.2 overlap https://nextcentury.atlassian.net/wiki/spaces/ITMC/pages/3476946948/Phase+I+Evaluation+Delegation+Materials). On this branch you can delete that entry from your db and then enter PID again after refreshing. You should see it select one that does not overlap with baseline.

nextcen-dgemoets (approved):
Got the following assignments in the instance mentioned in the PR.  They matched expectations based on the Confluence page.

adm-assignment

brianpippin (approved):
I am getting the expected ADM for my test cases.

phile-caci (approved):
I think we're good, ran through several text/surveys, checked all the baselines, and then spot checked a few least/most aligned assignments...

PR #175, opened 2024-11-25
query string edit so that phase 1 adms load in without have to reclick

dereknop:
https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?selectedIssue=ITM-728

http://localhost:3000/results

Before you had to click on 'Phase 1' again for the adms to load in. Now you should be able to select a scenario and have the adms popup without any additional clicking.


phile-caci (approved):
Nice.. it's fixed on my end.

PR #178, opened 2024-11-26
Itm 729 no email

kaitlyn-sharo:
http://localhost:3000/remote-text-survey?adeptQualtrix=true

Use the above link (logged in or out, should be available either way) to access the online-only experience.

Note: The instructions page will change. I have not received the text to put on it yet.

The online-only experience should start with an instructions page. At this point, no pid is assigned. As soon as "start" is pressed, the pid will be assigned. It should be impossible to get duplicate pids. Please try to break it. 

PIDs for online-only should start at 500, whereas email text participants should start at 300. 

The online-only experience should walk through all text scenarios and then head straight into the delegation survey with the correct adms chosen. When the delegation survey is finished, it should redirect to ourqualtrics.com?participant_id={pid}. This does not appear to be a real link and will likely change, but you should see the attempt to redirect and the failed access of the url.

The participant should appear in the log as "online" type.


node-graphql/server.schema.js
```diff
@@ -241,7 +241,7 @@ const typeDefs = gql`
     updateExperimenterUser(username: String, isExperimenter: Boolean): JSON,
     uploadSurveyResults(surveyId: String, results: JSON): JSON,
     uploadScenarioResults(results: [JSON]): JSON,
-    addNewParticipantToLog(participantData: JSON): JSON,
+    addNewParticipantToLog(participantData: JSON, lowPid: Int, highPid: Int): JSON,
```
dereknop, line 244: Are these needed? I was trying to track them down in the mutation and didn't find them.

dereknop (approved):
Everything worked as you described. PID generated was `202411500`. I see my results showing up 👍

node-graphql/server.schema.js
kaitlyn-sharo: I know past me knew they were needed, and I think present me agrees, but it looks like the addition of them got forgotten. The low one, at least, should replace 202411301 in line 581. Maybe we don't need high...thoughts?

dereknop: discussed on teams

dereknop:
Tested after `highPid` and `lowPid` were added into gql mutation. I already had online only results (pid: 5xx) and performed the race condition and the new pid generated for regular data collection was still in the 3xx's 👍

kaitlyn-sharo:
Now has progress table in it as well

kaitlyn-sharo:
> LGTM,
> 
> Redirect at the end (as expected)... the link looks live now. https://singuser67d7ec86.sjc1.qualtrics.com/jfe/form/SV_0pUd3RTN39qu9qS?participant_id=202411500&PROLIFIC_PID=null

Yeah, I forgot to update the description :)

PR #179, opened 2024-12-02
Itm 731 progress table

kaitlyn-sharo:
LOWER PRIORITY REVIEW:

This adds a new option for experimenters, evaluators, and admin to quickly see participant progress. It shows all the PIDs in the participant log (remember that there are no longer "extra" PIDs, in order for a PID to appear, it has been assigned to somebody somehow) along with how many sim responses, survey responses, and scenario responses they have completed. The dates are included. A column is dark green if it is a wrapped count, for example, are there 4 sim responses (all of them), 1 survey response, or 5 text responses. A column is light green if we have data for it. 

The cell coloring should transfer to the download. You will need to rerun the startup script for this because a new library was needed to get the styling. 

To find this new table, look in your user dropdown when logged in. Play around with the filters and make sure they work properly.

### DIFF
--- a/dashboard-ui/package.json
+++ b/dashboard-ui/package.json
@@ -62,7 +62,8 @@
         "survey-analytics": "^1.9.137",
         "survey-core": "^1.9.137",
         "survey-react-ui": "^1.9.137",
-        "typescript": "5.4.3"
+        "typescript": "5.4.3",
+        "xlsx-js-style": "1.2.0"
     },
     "scripts": {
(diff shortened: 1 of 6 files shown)


phile-caci (approved):
LGTM! Tested with an emailParticipant, text and delegation filled out as I completed them.

Awesome page to have!

brianpippin (approved):
Loaded and worked for me.

PR #181, opened 2024-12-02
fixed adm charts by evalNum

kaitlyn-sharo:
Before, the Adept ADM charts for eval 4 and 5 were the same because the code was not filtering by eval number (but the scenario names are the same for ph1 and dre). i added in the eval number filter (you'll have to rerun the startup script) and now all charts should work and be unique


dereknop (approved):
Awesome

brianpippin (approved):

PR #186, opened 2024-12-05
Itm 739

dereknop:
[ITM-739](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-739)

Adds new page http://localhost:3000/adm-probe-responses.

Rebuild graphql container first.

Check that the 4 different ADM tables are populating with the probe_ids and corresponding choice_ids. Also check that the tables download to excel okay in addition to the combined download. 

[ITM-739]: https://nextcentury.atlassian.net/browse/ITM-739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dereknop:
I added a fix onto this branch for the ADM Data page as well. Some of the adept scenarios for phase 1 were pulling adm runs from DRE. I added in a console log so you can check if you click on an adept phase 1 scenario run the object should show the proper eval. The query was previously not checking for evalNumber and the adept scenario names did not change, I believe that was the root of the problem.

phile-caci (commented):
Everything looks good, just noticed 2 things:

1. Data isn't appearing for the September Milestone tables, should they?
2. Not a big deal, but all the "get" request are being sent as POST, it doesn't break anything.

nextcen-dgemoets (commented):
Nice page!  Page function and downloads all looked good.  One neat little side benefit to this table is that in e.g. ADEPT's MJ5, you can visually see the branching narrative structure by seeing which probes have responses.

Not a huge deal, but in DRE `vol-dre-1-eval`, the oddly named probes `vol-dre-train2-Probe-11` and `vol-dre-train2-Probe-12` come *after* probe `4.10` in the scenario, so it would be nice if they appeared in the table after `4.10` as well.

Also, there are some `console.log` calls to remove in `resultsTable.jsx` and `admProbeResponses.jsx`.  :)

dereknop:
> Not a huge deal, but in DRE `vol-dre-1-eval`, the oddly named probes `vol-dre-train2-Probe-11` and `vol-dre-train2-Probe-12` come _after_ probe `4.10` in the scenario, so it would be nice if they appeared in the table after `4.10` as well.

That is lovely, I'll take a stab at changing that

dereknop:
> 1. Data isn't appearing for the September Milestone tables, should they?
Good point. I think I'll keep evals 3-5 and remove the others from the drop down because I don't think there would be any meaningful data at this point from then.
> 2. Not a big deal, but all the "get" request are being sent as POST, it doesn't break anything.
Weird, looking into this

dereknop:
> > Not a huge deal, but in DRE `vol-dre-1-eval`, the oddly named probes `vol-dre-train2-Probe-11` and `vol-dre-train2-Probe-12` come _after_ probe `4.10` in the scenario, so it would be nice if they appeared in the table after `4.10` as well.
> 
> That is lovely, I'll take a stab at changing that

It probably was way more effort than it was worth but this should all be fixed. Some of the adept ones were off too. I want to advocate for probe_ids and chocie_ids following a set format in the future.

kaitlyn-sharo (commented):
Neat page! When I click on a scenario, the first one I click doesn't work. Only subsequent scenarios show up. This doesn't seem to happen every time, but pretty often. Try refreshing the page, selecting one from Phase 1 (it will probably show up), then selecting Dry run, then selecting the first soartech one. 

More of a personal preference, but for consistency with the other pages, I think the download buttons should be more aligned right (i.e. align at the end of the table, rather than having a gap).

Also, usually our download buttons just say "Download", not "Download as Excel", but that _really_ doesn't matter, just something I noticed.

dereknop:
Ready for re-review

nextcen-dgemoets (approved):
All of the fixes looked good to me!

phile-caci (approved):
LGTM! didn't see any issues, clicked on most of the eval/scenarios,  all the downloads look right.

PR #193, opened 2024-12-10
Itm 755 skip text

kaitlyn-sharo:
Added new column to participant progress table (only for adept users and admin!). Shows "Copy Link" button when prolific id is available. Copies a link straight to your clipboard with the prolific id, contact id, pid, and other online parameters AND a new parameter called startSurvey. 

If startSurvey is true (which it will be when you copy the link), the participant will be brought directly to the survey (skipping the intro page and text). The pid should stay the same and data will not be overridden in the database.

The link should only be copiable if all 5 text scenarios have been completed!

### DIFF
--- a/dashboard-ui/src/components/Account/participantProgress.jsx
+++ b/dashboard-ui/src/components/Account/participantProgress.jsx
@@ -66,7 +66,7 @@ export function ParticipantProgressTable({ canViewProlific = false }) {
                 if (canViewProlific) {
                     obj['Prolific ID'] = res['prolificId'];
                     obj['Contact ID'] = res['contactId'];
-                    if (res['prolificId']) obj['Survey Link'] = `https://darpaitm.caci.com/remote-text-survey?adeptQualtrix=true&PROLIFIC_PID=${res['prolificId']}&ContactID=${res['contactId']}&pid=${pid}&class=Online&startSurvey=true`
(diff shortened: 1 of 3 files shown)


dereknop:
The copy link worked but it pointed to prod. be sure to replace that with local when testing

dereknop:
Worked great. I used my pid from the last pr that only had text scenarios completed. Replaced prod link with local host and took me straight to the survey. Upon completion, I was properly redirected and the contact id and prolific id were present. Progress table updated accordingly

dereknop (approved):

phile-caci (approved):
Worked for me too!

Just a QQ: Should I still get this link if the participant had already completed the survey?

dereknop:
Will try to resolve these conflicts

nextcen-dgemoets (approved):
LGTM.  I also tried a URL with bogus info (e.g., http://localhost:3000/remote-text-survey?adeptQualtrix=true&PROLIFIC_PID=foobar&ContactID=foobar&pid=1234&class=Online&startSurvey=true) and, well, I got through the survey and nothing bad seemed to happen in the progress table or the DB.

kaitlyn-sharo:
> Worked for me too!
> 
> Just a QQ: Should I still get this link if the participant had already completed the survey?

Good point. In the future, I'll remove the link if they've already completed the survey!

PR #195, opened 2024-12-11
fix mre to phase 1 crash

dereknop:
http://localhost:3000/humanProbeData

Bug: The page defaults to phase 1. If you select MRE and then return to phase 1, an undefined value causes a crash.

Test on dev, then please ensure it is no longer present on this branch. Oddly enough I tested on prod but didn't experience the crash but it happened every time for me locally.

### DIFF
--- a/dashboard-ui/src/components/AggregateResults/aggregateResults.jsx
+++ b/dashboard-ui/src/components/AggregateResults/aggregateResults.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import gql from "graphql-tag";
 import { useQuery } from '@apollo/react-hooks';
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets:
I'm still seeing this crash on the feature branch.   I rebuilt everything (`startup-scripts/dev_graphql_dashboard_ui.sh`):
itm-749

nextcen-dgemoets (approved):
LGTM.  I can switch among the three evaluation types willy-nilly to my heart's content.

phile-caci (approved):
LGTM!  No more errors switching evals on Sim Probe Data

I did get an error on the dev branch but none on this branch.

This is what's on dev:
image

I did check prod and got this(same line diff error, but looks like the fix should handle this also), prod was erroring on "desert", checked the new code, looks like it's handled:
image

PR #198, opened 2024-12-12
Adept online survey block navigation from page

dereknop:
For some mysterious reason, the Prompt component was not firing during the survey portion of the Adept online only experiment despite it working during the text-based scenarios. After trying to debug with Kaitlyn with no luck, I just decided to scrap that and do it a different way. 

Attempting to navigate away from the delegation survey with either the browser back button or the browser refresh button should give the user a warning instead of allowing them to navigate away immediately. The easiest way to test is to grab a link from the progress table to take you to an adept experiment delegation survey. Also make sure that it is working for the route OSU takes, which is using http://localhost:3000/survey.

Screenshot 2024-12-12 at 1 38 09 PM

The message should look like this for the back arrow. I couldn't edit the warning for reloading but it will still ask for confirmation with the user being able to hit cancel.

### DIFF
--- a/dashboard-ui/src/components/Survey/survey.jsx
+++ b/dashboard-ui/src/components/Survey/survey.jsx
@@ -726,7 +726,7 @@ class SurveyPage extends Component {
         window.addEventListener('popstate', this.handlePopState);
 
         // push initial state to prevent back navigation
-        window.history.pushState(null, '', window.location.pathname);
+        window.history.pushState(null, '', window.location.href);
     }
 
     handleBeforeUnload = (e) => {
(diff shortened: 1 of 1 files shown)


kaitlyn-sharo (approved):
The problem i saw initially is gone, where the query params mysteriously disappeared. I do always get the confirmation message when going back, but I have to confirm twice. As in, the first time, I get the normal pop up, press "Ok", then I have to press back again and then I get the browser popup which, when I press "Leave" takes me away. I'm okay with double checks before leaving, hence the approval.

phile-caci (approved):
Saw the same behavior that Kaitlyn got.  Have to hit the back button twice, but it pops up on all the pages I tried, which was most of them.

PR #199, opened 2024-12-12
Itm 757 legacy results

kaitlyn-sharo:
Added Legacy Results toggle to the survey results page. It is still organized in the new way with Blocks and DM numbers, but also includes Omnibus results. 

Additionally, I fixed the downloads so downloads should no longer contain columns that have no data. 

Finally, Version is the filter for legacy data and Eval is the filter for MRE/PH1 data. Filters should be erased when toggling between legacy and new data.

### DIFF
--- a/dashboard-ui/src/components/SurveyResults/resultsTable.css
+++ b/dashboard-ui/src/components/SurveyResults/resultsTable.css
@@ -179,3 +179,13 @@
 .small-column {
     min-width: 100px !important;
 }
+
+.simple-radios {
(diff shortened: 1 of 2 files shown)


phile-caci (approved):
LGTM!

Didn't see any issues toggling and filtering data
Downloads didn't have empty columns
Toggle between Legacy and DRE/Ph1 resets filters

nextcen-dgemoets (approved):
LGTM!  Didn't encounter any issues.

PR #201, opened 2024-12-12
Itm 758 del results definitions

kaitlyn-sharo:
Added definitions for legacy and MRE/DRE. I did not add definitions where the question was already written out with a comprehensible answer, only the ones that could cause some confusion. There are different definition tables/downloads for legacy and MRE/DRE. Also fixed some minor styling issues. 

This is the last planned delegation results PR, so if you see any other problems I missed with it, please mention it here!


nextcen-dgemoets (approved):
Overall, looks good.  Some minor things, but not sure if it's worth addressing them or not.
- Under `BX_DMY_Target`, it says "ADEPT 0.0-1.0", but that is for DRE.  For Phase 1, it was 0.2-0.8.  Maybe that should be indicated?
- At least on Windows/Acrobat, the Legacy definitions are kind of vertically cut off (see first pic).  The non-legacy definitions have these weird `?` characters in some of the headers (see second pic).  And both of them contain `\n` characters in the Likert scale descriptions (see first pic ).

definitions-legacy

definitions

kaitlyn-sharo:
> Overall, looks good. Some minor things, but not sure if it's worth addressing them or not.

I don't have much to do today, so I figured I'd give fixing these a shot. The Likert scales might still look a little weird because I just turned the \n's to white, so the spacing is a bit odd, but those are needed to make it look nice in the dashboard and I didn't want to have to keep track of two versions, so I think this is a good compromise. I vertically centered all the text so nothing gets cut off and I removed those special characters that must have come from copying and pasting headers from the dashboard

nextcen-dgemoets:
Looks great!  Like you say, the Likert Scales aren't lining up in the PDF the way they do in the table, but it's definitely an improvement and looks reasonable.  Thanks!

PR #203, opened 2024-12-13
Itm 761 sort progress

kaitlyn-sharo:
Updated progress table again. Added "hide columns" options. You can click the eyes next to each column to hide (easiest), or use the text box filter thing. The text box will update when you click the eyes so you can unhide columns using that. Hiding columns will also activate download filtered data, and downloading that will not have the hidden columns in the download. 

Added start and end times for delegation and survey (currently was showing end times, but was not labelled). 

Added some sorting options


nextcen-dgemoets (approved):
Looks really good!  I tried all the things and I couldn't even find anything to nitpick about!

dereknop (approved):
Looks awesome

PR #206, opened 2024-12-16
Text based scenarios navigation blocking

dereknop:
Standardized navigation blocking behavior for text-based scenarios to be the exact same as the delegation survey. 

Since we have two different entry points for the text-based scenarios, we should test both.

ADEPT entry:
http://localhost:3000/remote-text-survey?adeptQualtrix=true&ContactID=Derek&PROLIFIC_PID=314159&class=Online

OSU entry:
http://localhost:3000/login
Screenshot 2024-12-16 at 10 40 12 AM

You should be warned if you try to do any of the following before all of the text-based scenarios are complete:
- close the tab
- refresh the page
- hit the browser's back arrow

Upon completing OSU's entry point, it should allow you to close the tab with no warning. ADEPT's entry point should still redirect you as normal when the scenarios are complete.

### DIFF
--- a/dashboard-ui/src/components/Survey/survey.jsx
+++ b/dashboard-ui/src/components/Survey/survey.jsx
@@ -1,4 +1,4 @@
-import React, { Component, useEffect } from "react";
+import React, { Component, useEffect, useRef } from "react";
 import 'survey-core/defaultV2.min.css';
 import { Model } from 'survey-core';
 import { Survey, ReactQuestionFactory } from "survey-react-ui"
@@ -727,26 +727,7 @@ class SurveyPage extends Component {
 
     componentDidMount() {
         this.detectUserInfo();
-        
-        window.addEventListener('beforeunload', this.handleBeforeUnload);
-
-        // push initial state to prevent back navigation
-        window.history.pushState(null, '', window.location.href);
-        
-    }
-    
-    handleBeforeUnload = (e) => {
-        if (!this.state.surveyComplete) {
-            if (!window.confirm('Please finish the survey before leaving the page. If you leave now, you will need to start the survey over from the beginning.')) {
-                e.preventDefault();
-                e.returnValue = '';
-                return '';
-            }
-        }
-    };
-
-    componentWillUnmount() {
-        window.removeEventListener('beforeunload', this.handleBeforeUnload);
+       
     }
 
     detectUserInfo = () => {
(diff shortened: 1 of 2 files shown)


kaitlyn-sharo (commented):
Super weird edge case:
- Start the text survey (answer at least 1-2 questions)
- Press the browser back button
- Press OK to leave the page 
- The page is still up, which I had mentioned happened for the survey too, so normal behavior
- Press the browser back button
- Press cancel
- Now the PID is missing from the URL and therefore will not load the survey properly

brianpippin:
> > Super weird edge case:
> > 
> > * Start the text survey (answer at least 1-2 questions)
> > * Press the browser back button
> > * Press OK to leave the page
> > * The page is still up, which I had mentioned happened for the survey too, so normal behavior
> > * Press the browser back button
> > * Press cancel
> > * Now the PID is missing from the URL and therefore will not load the survey properly
> 
> There is similar, but slightly different behavior in the survey. I wonder if we want to keep pid in the store to avoid this? We can clear the pid out every time they hit a certain page, or maybe if the url is anything except remote-text-survey...just thinking out loud
> 
> Thinking more after testing, I think the pid is grabbed at the beginning of text and the beginning of survey, so if we just update the url right before we hit the survey page, nothing should go wrong. no need for the store or anything. As long as the pid is present in the url at the start of the survey, we should be good to go

In this case is the PID undefined?  So it they did finish everything and went back to ADEPT, it would show the PID as undefined?  Also do we still save anything if the PID is undefined?  We are missing some data from someone who said they did the text and survey from ADEPT.

kaitlyn-sharo:
> > > Super weird edge case:
> > > 
> > > * Start the text survey (answer at least 1-2 questions)
> > > * Press the browser back button
> > > * Press OK to leave the page
> > > * The page is still up, which I had mentioned happened for the survey too, so normal behavior
> > > * Press the browser back button
> > > * Press cancel
> > > * Now the PID is missing from the URL and therefore will not load the survey properly
> > 
> > 
> > There is similar, but slightly different behavior in the survey. I wonder if we want to keep pid in the store to avoid this? We can clear the pid out every time they hit a certain page, or maybe if the url is anything except remote-text-survey...just thinking out loud
> > Thinking more after testing, I think the pid is grabbed at the beginning of text and the beginning of survey, so if we just update the url right before we hit the survey page, nothing should go wrong. no need for the store or anything. As long as the pid is present in the url at the start of the survey, we should be good to go
> 
> In this case is the PID undefined? So it they did finish everything and went back to ADEPT, it would show the PID as undefined? Also do we still save anything if the PID is undefined? We are missing some data from someone who said they did the text and survey from ADEPT.

If the pid is lost from the url during the text portion, they are asked for their pid, where people may enter their prolific id. Then it will still not be in the URL but _should_ be in the url when redirected to adept. If the PID is undefined, the survey will never start, so nothing will be saved

dereknop:
Okay, thanks for the catch. Nobody else bother reviewing this while I sort this bug out.

dereknop:
Good to test

kaitlyn-sharo:
I'm still seeing that if I press back, then hit ok, my url changes but the page stays up, so eventually I get to the survey without a pid

dereknop:
okay... NOW it should be ft. Knox

kaitlyn-sharo (approved):
Using the adept route, I cannot break it! I am forced to stay on the page forever and cannot escape. 

Using the OSU route, I can remove the PID from the url, but as expected, everything still saves properly. The OSU route does not need the PID in the route since it does not redirect to the survey. Everything looks great to me

dereknop:
> I am forced to stay on the page forever and cannot escape.

giphy

nextcen-dgemoets (approved):
I could not break into Fort Knox.

PR #209, opened 2024-12-17
confirm emails OSU entry 

dereknop:
Adding in a bit more user proofing - Go to the OSU entry point for text scenarios and ensure it doesn't let you continue unless the emails match.

Screenshot 2024-12-17 at 12 29 01 PM

### DIFF
--- a/dashboard-ui/src/components/Account/login.jsx
+++ b/dashboard-ui/src/components/Account/login.jsx
@@ -25,7 +25,10 @@ class LoginApp extends React.Component {
             loginFailed: false,
             createAccountFailed: false,
             pidCreationFailed: false,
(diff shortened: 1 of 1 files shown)


kaitlyn-sharo (approved):
Exactly what I would have done...hopefully that's a good thing! :)

nextcen-dgemoets (approved):
Sorry I must've missed this PR.  Approved!  I like the use of `toLowerCase` and `trim` here.

nextcen-dgemoets:
> Sorry I must've missed this PR. Approved! I like the use of `toLowerCase` and `trim` here.

LOL I guess that's because it's from 45 minutes ago.

PR #210, opened 2024-12-17
Itm 770 protection

kaitlyn-sharo:
Before, anyone could change evaluator, admin, experimenter, and adept status using a simple query that only asked for a user's name and the updated status. Now, we check that the user calling the endpoint is an admin with a valid session id.

Session ids cannot be accessed from an endpoint and are therefore secure in our database. The admin page is password-protected now. When you enter your password to enter it, the page will get your session id and send it off with every move you make to make sure you are still a valid admin. 

To test, enter the admin page with your password, then change a few things, then make yourself not an admin and try to change something else. You should be kicked back to the password page. Once you fail the second time, you will be logged out and redirected to login. That login redirect will refresh the database and recognize you are no longer an admin. Nothing you tried should take effect.

Also, try using thunder client (or postman, or your favorite querier) to update admin/eval/adept/experimenter status using only the queries available to the public (i.e. you are not allowed to use free access to the database)

Note that this one definitely needs a graphql rebuild!


jaudick (approved):
Worked for me. I got sent to login on the second time i tried to change things after logging back in when I was no longer an admin

joshuale-caci (approved):
Worked for me as well and it sent me back after setting myself to admin and trying to change stuff two times.

nextcen-dgemoets:
> To test, enter the admin page with your password, then change a few things, then make yourself not an admin and try to change something else. You should be kicked back to the password page. Once you fail the second time, you will be logged out and redirected to login. That login redirect will refresh the database and recognize you are no longer an admin. Nothing you tried should take effect.

After I got kicked back to the password page, I entered my password and was let into the Admin page.  When I logged out, I couldn't get into the admin page.  Maybe the refreshing of the database took longer than expected?

kaitlyn-sharo:
> > To test, enter the admin page with your password, then change a few things, then make yourself not an admin and try to change something else. You should be kicked back to the password page. Once you fail the second time, you will be logged out and redirected to login. That login redirect will refresh the database and recognize you are no longer an admin. Nothing you tried should take effect.
> 
> After I got kicked back to the password page, I entered my password and was let into the Admin page. When I logged out, I couldn't get into the admin page. Maybe the refreshing of the database took longer than expected?

Did you fail twice on Admin?

nextcen-dgemoets:
Also, if everything's instantaneous, it kind of makes me feel like it should ask you to confirm removing yourself from the Admin group...maybe as a future enhancement.

nextcen-dgemoets:
> Did you fail twice on Admin?

Oh I see...I didn't realize that "fail the second time" meant I should try to make another admin change, get kicked out again, and enter my password-- then get kicked out.  Okay, that worked.  :)

Out of curiosity, why doesn't it just kick you out to the login page the first time it realizes you're not an admin?

kaitlyn-sharo:
> > Did you fail twice on Admin?
> 
> Oh I see...I didn't realize that "fail the second time" meant I should try to make another admin change, get kicked out again, and enter my password-- then get kicked out to login. Okay, that worked. :)
> 
> Out of curiosity, why doesn't it just kick you out to the login page the first time it realizes you're not an admin?

I have it set up for failing twice in case a session becomes invalid for some reason. Entering your password will get the session fresh. But then if your action (updating a user status) fails again, it decides something is wrong so it kicks you out for good, sending you back to login. Now when you have to re-log-in, it will know you are not an admin and won't let you on the page

phile-caci (approved):
Perfect!  I was able to do this before but not after the changes:

image

nextcen-dgemoets (commented):
Well, I tested the first two parts.  I didn't test with a querier, but @phile-caci did, and I'd have to set that up and we need to move to dev-to-main...

PR #211, opened 2024-12-17
add blocks for general users

kaitlyn-sharo:
One user took the delegation survey instead of going in through the text scenario email route. This prevents that in two ways:

1. If they go in through the participantText link like we expect, create account and sign in are no longer there
2. If they create an account, since they are not an admin, evaluator, or experimenter, the drop down to take the surveys are no longer there

Routes anyone can access:
/login
/participantText
/reset-password/:token
/myaccount
/remote-text-survey

Routes only elevated users (admin, evaluator, experimenter) can access (others will be redirected to home)
/results
/adm-results
/adm-probe-responses 
/humanSimParticipant 
/scenarios
/admin
/participant-progress-table
/pid-lookup
/survey
/survey-results
/review-text-based
/review-delegation
/text-based
/text-based-results
/humanProbeData
/human-results
/dre-results/rq1
/dre-results/rq2
/dre-results/rq3
/dre-results/exploratory-analysis
/participantTextTester - will redirect to /login

### DIFF
--- a/dashboard-ui/src/components/Account/login.jsx
+++ b/dashboard-ui/src/components/Account/login.jsx
@@ -223,7 +223,7 @@ class LoginApp extends React.Component {
                 <div className="row justify-content-center align-items-center h-100 center-container">
                     <div className="login-container">
                         <Tabs className='p-1' defaultActiveKey={this.props.isParticipant ? 2 : 1}>
-                            {this.props.history.location.pathname != '/participantTextTester' &&
+                            {this.props.history.location.pathname != '/participantTextTester' && this.props.history.location.pathname != '/participantText' &&
                                 <Tab eventKey={0} title="Create Account">
                                     <div>
                                         <div className="sign-in-instructions">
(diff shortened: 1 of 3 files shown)


nextcen-dgemoets (approved):
Entering through the participantText link, "create account" and "sign in" were not there.
When I created a new account, it directed me to the (production) participantText link, and none of the other links were there:  only Home, My Account, and Logout.

No more silly users!

jaudick (approved):

nextcen-dgemoets (approved):
Everything looks good.  An evaluator user couldn't get to /admin or /pid-lookup (which is in the list in the PR description), but I believe that's as intended.

kaitlyn-sharo:
> Everything looks good. An evaluator user couldn't get to /admin or /pid-lookup (which is in the list in the PR description), but I believe that's as intended.

Ah yes, I was not very specific about which elevated users could do which things

PR #212, opened 2024-12-18
Dev to Main

brianpippin:


nextcen-dgemoets:
I'm having a problem (first found on `graphql-harden` but reproduced here prior to it being merged into `main`) where if I start taking the delegation survey, then try to navigate away, I get the OK/Cancel popup, but the OK button doesn't do anything (just redisplays the pop-up) so I can't get to another page.  I eventually found a way out, but I think the OK button should work, no?

kaitlyn-sharo:
> I'm having a problem (first found on `graphql-harden` but reproduced here prior to it being merged into `main`) where if I start taking the delegation survey, then try to navigate away, I get the OK/Cancel popup, but the OK button doesn't do anything (just redisplays the pop-up) so I can't get to another page. I eventually found a way out, but I think the OK button should work, no?

No, that was derek's "trap-user" PR. At least for now, this is the intended behavior to prevent users from quitting accidentally. You truly get trapped on all these pages. The only way out is to change the URL

kaitlyn-sharo (approved):
Nothing seemed broken

nextcen-dgemoets:
> The only way out is to change the URL.

While the OK/Cancel dialog was up, I was able to get out by closing the browser tab.

nextcen-dgemoets (approved):
I haven't run into any issues beyond what I reported in the OK/Cancel dialog, which probably isn't an issue.

phile-caci (approved):
Not seeing any issues either, clicked all around, went through an online ADEPT survey, no more graphql end points, retested updateAdminUser without being logged in, not allowed as expected.

PR #214, opened 2024-12-18
minor progress table updates

kaitlyn-sharo:
There was a bug in the progress table where the evaluation was not being grabbed from the delegation survey properly (see prod participant 330). In the finding of this bug, I remembered that I wanted a search-by-pid feature on this page to make things easier to quickly look through, so I added that, too.


jaudick (approved):
Worked for me

nextcen-dgemoets (approved):
Looks good.  It even works if PID is hidden, and the filter applies to the download.

PR #215, opened 2024-12-18
Itm 772 admin approval

kaitlyn-sharo:
Rebuild first, and run the 772 ingest PR

We do not want random people creating accounts. This ticket sends new users to a waiting room, while adding their information to a new table in the admin page. Admins can access the table and approve or reject prospective users. In the process, they can also choose the user's permissions (admin, evaluator, experimenter, adept).

If a user is rejected, their waiting room status will be updated accordingly and they should no longer appear in the admin table.

If a user is approved, they will be sent to the home page (no longer able to access /awaitingApproval) and they should no longer appear in the admin table.

Rejected users and users awaiting approval should not be able to access _anything_ in the dashboard. The only action they should be able to take is returning to the login page.

### DIFF
--- a/dashboard-ui/src/components/Account/adminPage.jsx
+++ b/dashboard-ui/src/components/Account/adminPage.jsx
@@ -46,6 +46,12 @@ const UPDATE_ADEPT_USER = gql`
     }
 `;
 
+const UPDATE_USER_APPROVAL = gql`
+    mutation updateUserApproval($caller: JSON!, $username: String!, $isApproved: Boolean!, $isRejected: Boolean!, $isAdmin: Boolean!, $isEvaluator: Boolean!, $isExperimenter: Boolean!, $isAdeptUser: Boolean!) {
+        updateUserApproval(caller: $caller, username: $username, isApproved: $isApproved, isRejected: $isRejected, isAdmin: $isAdmin, isEvaluator: $isEvaluator, isExperimenter: $isExperimenter, isAdeptUser: $isAdeptUser)
+    }
+`;
+
 const GET_CURRENT_SURVEY_VERSION = gql`
     query getCurrentSurveyVersion {
         getCurrentSurveyVersion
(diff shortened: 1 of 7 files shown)


dereknop (approved):
Worked great, this is really cool.

After trying to create an account:
Screenshot 2024-12-19 at 12 28 02 PM

Logged in as admin:
Screenshot 2024-12-19 at 12 28 59 PM

After denying the user and then trying to sign in as that user:
Screenshot 2024-12-19 at 12 29 43 PM

Went back through and approved the user with different user roles and it was all reflected in mongo as well as UI.

nextcen-dgemoets:
Should we require the admin assign at least one role to the user when accepting?  Maybe it's way not to reject/ban a user, but figure out what role they should have?  Not sure...

nextcen-dgemoets:
> Should we require the admin assign at least one role to the user when accepting? Maybe it's way not to reject/ban a user, but figure out what role they should have? Not sure...

@brianpippin  Any thoughts on this?  Keep it how it is, require an assignment, or write a ticket for it?

brianpippin:
> > Should we require the admin assign at least one role to the user when accepting? Maybe it's way not to reject/ban a user, but figure out what role they should have? Not sure...
> 
> @brianpippin Any thoughts on this? Keep it how it is, require an assignment, or write a ticket for it?

I don't think we need to have the admin assign the user a role, they probably will when approving but I don't think we make it mandatory.  They can easily use the tools after approving to assign them a role, but I don't think we should hold up the ticket for it.  If it becomes an issue we can make a ticket to add it.

nextcen-dgemoets (approved):
> If you believe this is a mistake, please contact an administrator.

Should we indicate where/how/who, like with an email address?  Or maybe if there is a legitimate mistake, the person should have an idea whom to contact (somebody in their organization) and if it's not a mistake (it's a random person), then we don't want to divulge any information.  I think that works.

kaitlyn-sharo:
> > If you believe this is a mistake, please contact an administrator.
> 
> Should we indicate where/how/who, like with an email address? Or maybe if there is a legitimate mistake, the person should have an idea whom to contact (somebody in their organization) and if it's not a mistake (it's a random person), then we don't want to divulge any information. I think that works.

Yes, Brian and Jennifer had said that anyone who reaches this page will know who to contact if there has been a real mistake, and they will let us know when they have made an account too

PR #216, opened 2024-12-19
Itm 777

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-777)

After asking Brian, this ticket turned into asking for confirmation before changing anyone's admin privileges instead of just removing it from yourself. 

http://localhost:3000/admin

Test that you get prompted to confirm before admin privileges are added or removed from a user.


nextcen-dgemoets:
I'm finding this very important message pretty illegible.  Is this what it looks like for everybody?  Did I miss the "dark theme" option?
confirm-admin

nextcen-dgemoets:
Also, maybe the list of users should also include the email address, just like they do in the user lists.  I'm not sure if I want to remove admin privileges from Derek, but I *definitely* know I want to remove them from derek.noppinger@caci.com!

kaitlyn-sharo (commented):
Functionally, this works great. I do agree with both of Darren's comments (it would be nice to have the text darker and have emails in parentheses next to usernames).

Additionally, it would be great if you pulled the confirm modal out of the InputBox component, because right now that's being duplicated for all 4 role boxes, but only used for one. You should be able to pull the modal out into the top level and then send the setShowConfirmation function in as a parameter. It should work the same, but keep the modal outside of the component where I don't think it necessarily belongs. I could be convinced otherwise

nextcen-dgemoets (approved):
Yay!  Looks a lot better and works great, too.

kaitlyn-sharo (approved):
Visually, it looks way better. I personally still think there's too much admin-specific code in the input box component, which should be a short, simple component, but it isn't affecting the functionality, so I don't want to delay the merge

PR #218, opened 2024-12-20
HIGH PRIORITY: Itm 782 competence

kaitlyn-sharo:
Added Competence Error to the RQ 1/3/4 tables.

Also fixed some missing data in these tables due to mismatching delegation materials against the participant log. 

To test, make sure the table still loads (including DRE, which should not have changed except an empty column for competence).

Check against prod to see that certain pids (i.e. 302) are showing up in this new version


joshuale-caci (approved):
I see the tables populate fine and including dre also works. I don't have the updated database yet but it looks nominal on my side.

phile-caci (approved):
LGTM, everything still works, RQ 1/3/4 has new Competence Error column and is populated.

jaudick (approved):
Appears for me

dereknop (approved):
I experienced the same issue with the dre toggle but for this weekend, all of the important data appears for me

nextcen-dgemoets (approved):
After the fact, but everything looks good on my end.  302 was present, but not on prod.  The DRE toggle worked for me before running the ingest script (043), but broke thereafter...which everybody knows because the problem was already diagnosed, fixed, and merged in the time for me to run the weekly script...

PR #220, opened 2024-12-23
Keep progress table sorted on refresh

kaitlyn-sharo:
Before, the progress table would revert to default sorting when "Refresh Data" was pressed. To test this, sort the table some way other than default, and in another tab start a text survey (you only have to get as far as to be assigned a PID). Press refresh data and see your sort choice remains in the filter options, but the table is now unsorted.

This is fixed in this PR. So try the same steps above, and you should see the sorting remains stable on refresh. Additionally, I added a "fetching data" spinner because I've noticed on prod, sometimes the data comes in batches and can take a second. This makes it more clear when the data has fully been re-fetched. Filtering and sorting should remain steady through this.


joshuale-caci (approved):
My sort by stayed the same when I refreshed the data and the addition of the spinner is nice!

PR #223, opened 2025-01-02
fixed rq5 table ST vs AD

kaitlyn-sharo:
RQ5 was only looking for DryRunEval for adept scenarios in order to decide between ST and AD, affecting the TA1 column and scenario column. This adds in the "adept" check for phase 1 adept scenarios.

Fixed RQ6 in the same way


nextcen-dgemoets (approved):
Huh, never noticed that.

Looks good-- on dev, there is no ADEPT for RQ5 and RQ6, but on the feature branch, there is!

dereknop (approved):
> on dev, there is no ADEPT for RQ5 and RQ6, but on the feature branch, there is!

ditto 👍

PR #229, opened 2025-01-09
Itm 814 update rq134

kaitlyn-sharo:
1. Added a reset filters button (does not change the DRE checkbox and does not unhide hidden columns. That is intentional). To see it in action, add some filters and you'll see it pop up next to "Showing x of y columns". Click it and see the table reset
2. Moved delegator_id to the far left of the table. It just makes sense.
3. Added hide columns feature
4. Added search-by-pid feature
5. Moved filters around to make it more visually pleasing (also changed it so that only 2 tags will appear)
6. BUG FIX: Prior to this ticket, if you checked the Include DRE checkbox and then navigated to the DRE data, all data would be duplicated. That is fixed here by resetting the DRE checkbox every time evalNumber updates.

### DIFF
--- a/dashboard-ui/src/components/Research/tables/rq1-rq3.jsx
+++ b/dashboard-ui/src/components/Research/tables/rq1-rq3.jsx
@@ -204,7 +204,6 @@ export function RQ13({ evalNum, tableTitle }) {
                     multiple
                     options={ta1s}
                     value={ta1Filters}
-                    filterSelectedOptions
                     size="small"
                     limitTags={2}
                     renderInput={(params) => (
@@ -219,7 +218,6 @@ export function RQ13({ evalNum, tableTitle }) {
                 <Autocomplete
                     multiple
                     options={ta2s}
-                    filterSelectedOptions
                     value={ta2Filters}
                     size="small"
                     limitTags={2}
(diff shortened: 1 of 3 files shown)


dereknop (approved):
Search for PID is awesome. DRE data not duplicating. Delegator_ID at far left of tables. Reset filters worked as expected. Awesome work. For some reason, I thought the hide columns feature was already live. It works as expected but appeared like this for me:
Screenshot 2025-01-10 at 10 03 27 AM

Sometimes it overlapped with the column header a little bit. Small detail, feel free to ignore.

nextcen-dgemoets (approved):
So much fun stuff!  Really great-- a pleasure to use.

Would there be room to widen the "Hidden Cols" dropdown?  It might better accommodate the column names in the main section, plus some of the labels in the dropdown itself.  But it's also nice having all of the dropdowns the same uniform width, so I don't feel strongly about it.
clip

nextcen-dgemoets:
> Added some bottom padding to address the eye issue, and moved the filters around some more to make Hidden Columns bigger (should scale relatively well) 

I like the wider dropdown, thanks!

kaitlyn-sharo:
Got an okay on Derek concerning the eyes on teams. Merging!

PR #230, opened 2025-01-10
Itm 821 adept dre data

kaitlyn-sharo:
Added more data to RQ 1/3/4. 

Yes, variables will need to be updated. I plan on doing that in a separate ticket maybe today, maybe Monday. 

To test, run https://github.com/NextCenturyCorporation/itm-ingest/pull/102

Then check the RQ 1/3/4 tables. You'll notice a few new columns. Since I haven't updated variables yet, I'll do my best to explain here (new columns will only show up if the drop down is PH1):
**DRE_ADM|Target** - The ADM probes were run through the DRE server. This is the alignment against the target specified in the Target column. If Include DRE Data is selected, this still holds true: the number in this column will be the value gotten from running the ADM probes through the DRE server, meaning they should match the Alignment score (ADM|target) column if the DRE dropdown was selected. The Alignment score (ADM|target) if the PH1 dropdown is selected is now ALWAYS the ADM alignment against the target when run through the PH1 server

Everything else follows this pattern, so I won't over-explain:
**DRE_Delegator|Target** - The text probe alignment from the delegator according to the DRE server. Alignment score (Delegator|target) with PH1 dropdown is text probe alignment from delegator according to PH1 server
**DRE_ADM_loading** - We looked at the most and least aligned targets for the delegator (text probes) according to the DRE server. If the most aligned target matched what was shown as aligned, or least aligned matched what was shown as misaligned, this is "normal". Otherwise, it's "exception". Baseline should always be normal.
**DRE_Delegator|Observed_ADM** - The comparison value between text probes and the delegation ADM (fewer probes) according to the DRE server

### DIFF
--- a/dashboard-ui/src/components/Research/tables/rq1-rq3.jsx
+++ b/dashboard-ui/src/components/Research/tables/rq1-rq3.jsx
@@ -204,7 +204,6 @@ export function RQ13({ evalNum, tableTitle }) {
                     multiple
                     options={ta1s}
                     value={ta1Filters}
-                    filterSelectedOptions
                     size="small"
                     limitTags={2}
                     renderInput={(params) => (
@@ -219,7 +218,6 @@ export function RQ13({ evalNum, tableTitle }) {
                 <Autocomplete
                     multiple
                     options={ta2s}
-                    filterSelectedOptions
                     value={ta2Filters}
                     size="small"
                     limitTags={2}
@@ -236,7 +234,6 @@ export function RQ13({ evalNum, tableTitle }) {
                     multiple
                     options={scenarios}
                     value={scenarioFilters}
-                    filterSelectedOptions
                     size="small"
                     limitTags={2}
                     renderInput={(params) => (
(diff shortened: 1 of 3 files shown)


kaitlyn-sharo:
> The only silly comment I have, I guess, relates to the column names and/or order in the chart. It was nice to have the three "Alignment score" columns next to each other, but now they are broken up by the DRE-related columns. But since I believe the new columns are also alignment scores, should be they be `Alignment score (DRE_ADM|Target)` and `Alignment score (DRE_Delegator|Target)`? Would that be more consistent?


ADEPT/Jennifer had mentioned these variables, though it was probably just to give me an idea quickly of what the columns were supposed to be. I also like the consistency, so I will change it. Good thing I hadn't uploaded my new variable document to the other ticket yet!!

kaitlyn-sharo:
Updated to DRE [old variable name], as the ADMs/Delegators are not from DRE, which is how Alignment score (DRE_ADM|target) makes it sound, but instead the alignment score is from the DRE server

PR #234, opened 2025-01-14
added kdmas to adm-probe-responses for ADEPT

kaitlyn-sharo:

KDMAs should show up for ADEPT only. KDMA should match the target attribute. Should appear in both downloads.

Styling update is because the evaluation menu had a weird size on a smaller screen. This fixes it for a normal-size screen, but not for mobile/tablet. I will be making a ticket to put in a more long-term fix for that styling bug


nextcen-dgemoets (approved):
LGTM!

dereknop (approved):
good stuff

PR #236, opened 2025-01-15
Itm 737

dereknop:
This pr cuts down the obnoxiously long build time for the dashboard-ui container. There is room for improvement here but the build time should now be about 1/2 of what it was before. To test make sure you are starting from scratch: 
`docker system prune -a --volumes`

Then build the UI as normally:
```
docker build --tag dashboard-graphql node-graphql/.
docker build --tag dashboard-ui dashboard-ui/.

cd docker_setup
docker-compose -f docker-compose-dev.yml up -d
```

Before this pr my local build time was nearly 600 seconds, that is down to about 240 seconds for me.


kaitlyn-sharo:
I could be wrong, but if you really want to start fresh, you will have to stop those containers as well, because pruning will only remove stopped containers. docker stop $(docker ps -a -q) will stop all containers. 

This would also mean you'd lose any other docker containers you have running, so you may want to do it container-by-container (or image-by-image) either in the command line or docker desktop. I'm also guessing we'll need to refresh our mongo db after.

kaitlyn-sharo (approved):
Step 1: 32.9 s
Step 2: 199.3 s
Step 3: 11.1 s

nextcen-dgemoets (approved):
It was shorter overall:

On `main`:
Step 1: 38.8 s
Step 2: 433.5 s
Step 3: 22 s
Total:  494.3s

On `itm-737`:
Step 1: 41.8 s
Step 2: 242.4s
Step 3: 31s
Total:  315.2s

brianpippin:
I was about to say before this gets merged we should make sure all the functionality is in tact and not just inspect the load times, but we can double check all of that on dev.

kaitlyn-sharo:
> I was about to say before this gets merged we should make sure all the functionality is in tact and not just inspect the load times, but we can double check all of that on dev.

I did take a look around the dashboard and checked the console for errors. I also checked to make sure npm i and run test worked outside of docker. Everything I checked was fine, but I was not thorough

PR #238, opened 2025-01-17
added treatment counts to adms for adept scenarios

kaitlyn-sharo:
http://localhost:3000/results

This page now contains a "current count" parenthetical next to treatments. The number next to "current count" should be the total number of treatments applied up to and including the action (it should match treatments_applied).

I clicked through many of the evals/scenarios/targets to make sure there weren't any undefined errors, but please do the same and make sure the counts show for all treatments across ADEPT and ST.

### DIFF
--- a/dashboard-ui/src/components/Results/resultsTable.jsx
+++ b/dashboard-ui/src/components/Results/resultsTable.jsx
@@ -334,7 +334,7 @@ class ResultsTable extends React.Component {
                                                                 <Table className='itm-table' stickyHeader aria-label="simple table">
                                                                     <TableBody className='TableBodyScrollable'>
                                                                         {testData.history.map((item, index) => (
-                                                                            <ActionRow key={item.command + index} item={item} ta1={this.formatScenarioString(this.state.scenario).split(':')[0]} />
+                                                                            <ActionRow key={item.command + index} item={item} />
                                                                         ))}
                                                                     </TableBody>
                                                                 </Table>
@@ -377,11 +377,11 @@ class ResultsTable extends React.Component {
 }
 
 
-function ActionRow({ item, ta1 }) {
+function ActionRow({ item }) {
     const [open, setOpen] = React.useState(false);
 
     const renderNestedItems = (item, response = null) => {
-    // pass response through for ADEPT treatment counts
+        // pass response through for treatment counts
         if (isObject(item)) {
             return renderNestedTable(item, response);
         } else if (Array.isArray(item)) {
@@ -403,7 +403,7 @@ function ActionRow({ item, ta1 }) {
             <Table size="small">
                 <TableBody>
                     {Object.entries(tableData).map(([key, value], i) => {
-                        if (isTreatment && response && key == 'treatment' && value == 'Hemostatic gauze') {
+                        if (isTreatment && response && key == 'treatment') {
                             for (const c of (response?.characters ?? [])) {
                                 if (c['id'] == character) {
                                     for (const injury of c['injuries']) {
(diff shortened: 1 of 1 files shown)


dereknop (approved):
Looks good. Saw the counts for ADEPT in DRE and P1 (intent actions skipped over). Only present for adept and not ST.

brianpippin:
Looking through the code I have one concern.  What if in the future ST decided to use something similar and we have this code in here about ADEPT and then forget about it.  Would it make more sense to only show it if available and not tie it to one TA just so we don't run into any issues down the road?

nextcen-dgemoets:
> Looking through the code I have one concern. What if in the future ST decided to use something similar and we have this code in here about ADEPT and then forget about it. Would it make more sense to only show it if available and not tie it to one TA just so we don't run into any issues down the road?

With no regard to any impact on the code, this does sound like a good idea.  Also, in theory the code doesn't have to call out hemostatic gauze explicitly:  pressure bandages in particular could be a good candidate for a quantized injury, even if they weren't used as such in the P1E.  This could also have the side effect (side benefit?) of showing over-treatment.  If an ADM tries to apply two tourniquets to a leg, the second time you'd see "(current count: 2)"...where "count" is "treatment count" not necessarily the number of supplies consumed.

kaitlyn-sharo:
Finishing up another PR and then I will make these changes, showing treatment counts for all treatments and all TA1s

nextcen-dgemoets:
@brianpippin One note about this, is that it will only increment the current count for *successful* treatments, for better or for worse.  A tourniquet on the neck will not get `(current count: 1)`.  Nor will `Nasopharyngeal airway` to the `head`, because the injury was in the `right face`.  So just FYI that this is also a way to see what treatments are "unsuccessful" in that they don't treat an injury.  A corollary to this is that, e.g., `Pain Medications` should never show a current count, because it doesn't treat an injury.

If we don't like this, then probably the dashboard would have to keep track of the number of `APPLY_TREATMENT` actions with the same treatment/location.

nextcen-dgemoets (approved):
Looks good-- didn't encounter any problems.  We should just make sure this is the behavior we want re: successful treatments, as noted in my previous comment.

dereknop (approved):

kaitlyn-sharo:
> @brianpippin One note about this, is that it will only increment the current count for _successful_ treatments, for better or for worse. A tourniquet on the neck will not get `(current count: 1)`. Nor will `Nasopharyngeal airway` to the `head`, because the injury was in the `right face`. So just FYI that this is also a way to see what treatments are "unsuccessful" in that they don't treat an injury. A corollary to this is that, e.g., `Pain Medications` should never show a current count, because it doesn't treat an injury.
> 
> If we don't like this, then probably the dashboard would have to keep track of the number of `APPLY_TREATMENT` actions with the same treatment/location.

@brianpippin any opinions on this? Been holding off on the merge for confirmation that what I have is fine, or a recommendation to change.

brianpippin:
> > @brianpippin One note about this, is that it will only increment the current count for _successful_ treatments, for better or for worse. A tourniquet on the neck will not get `(current count: 1)`. Nor will `Nasopharyngeal airway` to the `head`, because the injury was in the `right face`. So just FYI that this is also a way to see what treatments are "unsuccessful" in that they don't treat an injury. A corollary to this is that, e.g., `Pain Medications` should never show a current count, because it doesn't treat an injury.
> > If we don't like this, then probably the dashboard would have to keep track of the number of `APPLY_TREATMENT` actions with the same treatment/location.
> 
> @brianpippin any opinions on this? Been holding off on the merge for confirmation that what I have is fine, or a recommendation to change.

I think what you have is fine to merge for now, but let's create another ticket to investigate and determine how we want to handle Darren's comment.  I think a conversation with Jennifer included will be necessary.

kaitlyn-sharo:
> > > @brianpippin One note about this, is that it will only increment the current count for _successful_ treatments, for better or for worse. A tourniquet on the neck will not get `(current count: 1)`. Nor will `Nasopharyngeal airway` to the `head`, because the injury was in the `right face`. So just FYI that this is also a way to see what treatments are "unsuccessful" in that they don't treat an injury. A corollary to this is that, e.g., `Pain Medications` should never show a current count, because it doesn't treat an injury.
> > > If we don't like this, then probably the dashboard would have to keep track of the number of `APPLY_TREATMENT` actions with the same treatment/location.
> > 
> > 
> > @brianpippin any opinions on this? Been holding off on the merge for confirmation that what I have is fine, or a recommendation to change.
> 
> I think what you have is fine to merge for now, but let's create another ticket to investigate and determine how we want to handle Darren's comment. I think a conversation with Jennifer included will be necessary.

Ticket added.

PR #239, opened 2025-01-21
Itm 779 Phase 1 Home Page

kaitlyn-sharo:
Last graph is a match with the correct data
image

https://nextcentury.atlassian.net/wiki/spaces/ITMC/pages/3495493634/Phase+I+Evaluation+Results

Graphs should match the latest powerpoints on this Confluence page. The RQ3.2 graphs should be the only ones missing, as Javascript has no method for making regressions. 

DRE should not have changed (aside from adding the accordions and SkipTo menu in the bottom left). Also the RQ1 scatters are now starting at 0 based on feedback from other teams. You can check this against Prod.

PH1 should be the default page and contain several new graphs (that should, again, not appear on DRE) to match the powerpoints. Some graphs include logic changes and data changes from DRE based on new calculation methods. This should all be able to be checked visually when comparing against the confluence page. 

Switching between evals should still work.

One set of text on the RQ3 page was fixed after a conversation with Jennifer.

Code was organized better so others can more easily edit in the future.

UX features mentioned above were added for easier navigation on the page.


nextcen-dgemoets:
So far so good...but I'm having a heck of a time finding the corresponding graphs in the PowerPoints compared to the dashboard.  Will give it another go tomorrow.

kaitlyn-sharo:
> So far so good...but I'm having a heck of a time finding the corresponding graphs in the PowerPoints compared to the dashboard. Will give it another go tomorrow.

One of them for RQ1 looks a little different, where in the powerpoint, it is an average of the two attributes, but on the home page, it shows points for both attributes (it's got these colorful circles and lines). Jennifer said to keep it in this format because it is showing the same thing. The rest should be identical (except the colors)

dereknop (approved):

nextcen-dgemoets:
- Is there any way to make the label in some of the tooltips (before the colon) a little brighter?  The value to the right of the tooltip is black, but the left part is *very* dim-- a light, light gray.
- Some of the little squares inside the large rectangles in the RQ1.2 graphs seem different, e.g. ADEPT's Kitware-Delegator|target-misaligned.  They are also laid out differently, but that's not a problem.
  - Actually, as I look more closely, maybe it's just that the little square is smaller in the PPT than the dashboard.
  - And actually, I was looking at the 1/17 version of the slides, not the 1/20 version.
  - But in the 1/20 version, I don't see as many graphs as in the dashboard.  But from what I can see, it looks good...
- In RQ2.1 what are the values in the x-axis?  They aren't in the PPT versions.  It looks like it goes from -0.5 to 1.5?
- In RQ2.2, Parallax-SoarTech group targets, the Aligned ADM seems to missing the variability range bars-- they are in the PPT.  The equivalent Kitware also lacks the variability range bars, but this matches the PPT.
- In the RQ2.3 line graphs, in the PPT it's plotted on a grid where the distance between e.g. 0.2 and 0.3 is closer to the same on each axis compared to the dashboard.  I think this makes differences look more pronounced/noticeable.  But this is just a nit and maybe it wasn't possible to replicate the same way on the dashboard, e.g. when printing the (long) ST alignment target IDs.
- In RQ3.1, is it possible/desirable to express these as percentages like in the PPT, possibly with significant figures?  e.g., 0.82608696 would go to 82.6.  Obviously not a big deal.
- I'm sure it's me, but I'm not seeing the "Alignment scores between delegator and observed ADMs by Attribute" in the PPT, and I'm not seeing the "Alignment score as predictor of delegation choice" in the dashboard.

kaitlyn-sharo:
> * In RQ1, is there any way to make the label in some of the tooltips (before the colon) a little brighter?  The value to the right of the tooltip is black, but the left part is _very_ dim-- a light, light gray.
**Updated all tooltips that were light colors (there was one in RQ3 as well). Thanks for catching this.**

> * Some of the little squares inside the large rectangles in the RQ1.2 graphs seem different, e.g. ADEPT's Kitware-Delegator|target-misaligned.  They are also laid out differently, but that's not a problem.
>   
>   * Actually, as I look more closely, maybe it's just that the little square is smaller in the PPT than the dashboard.
>   * And actually, I was looking at the 1/17 version of the slides, not the 1/20 version.
>   * But in the 1/20 version, I don't see as many graphs as in the dashboard...maybe this is what you were saying in this morning's comment above.  But from what I can see, it looks good...
**This is the same as DRE and Brian and Jennifer are good with it**

> * In RQ2.1 what are the values in the x-axis?  They aren't in the PPT versions.  It looks like it goes from -0.5 to 1.5?
**Updated this to hide those numbers. That axis is required to get everything lined up, but can be hidden. I just didn't notice.**

> * In RQ2.2, Parallax-SoarTech group targets, the Aligned ADM seems to missing the variability range bars-- they are in the PPT.  The equivalent Kitware also lacks the variability range bars, but this matches the PPT.
**Good catch. The bars aren't missing, it's just that there's no variability. I am checking with Jennifer on this one because in the PPT, her Parallax-ST has a variability that goes above 1, so I'm just checking if this was a image problem or if I need to look into it further.**

> * In the RQ2.3 line graphs, in the PPT it's plotted on a grid where the distance between e.g. 0.2 and 0.3 is closer to the same on each axis compared to the dashboard.  I think this makes differences look more pronounced/noticeable.  But this is just a nit and maybe it wasn't possible to replicate the same way on the dashboard, e.g. when printing the (long) ST alignment target IDs.
**Checked with Brian and he says this looks fine as-is**

> * In RQ3.1, is it possible/desirable to express these as percentages like in the PPT, possibly with significant figures?  e.g., 0.82608696 would go to 82.6.  Obviously not a big deal.
**Brian and Jennifer are okay with this one**

> * I'm sure it's me, but I'm not seeing the "Alignment scores between delegator and observed ADMs by Attribute" in the PPT, and I'm not seeing the "Alignment score as predictor of delegation choice" in the dashboard.
As I mentioned in the description, the Alignment score as predictor of delegation choice cannot be done in the dashboard. JavaScript does not have that functionality. I also mentioned an extra graph or two. The other one you mentioned is one of those. We used it for DRE and I believe Jennifer said to just leave them for PH1.

brianpippin (approved):

kaitlyn-sharo:
@nextcen-dgemoets I spoke with Jennifer about the RQ2 graph variability problem. She confirmed that she has 0 variability for both Parallax and Kitware on Aligned ST group targets. Her graph was wrong, what I have is correct

nextcen-dgemoets (approved):

PR #241, opened 2025-01-24
New experiment changes

dereknop:
Survey shouldn't have the location of treatments for SoarTech scenarios anymore. Eval number should be set to 6 with a eval name of 'Jan 2025 Eval'. New pids should start at 202501700. Take the OSU route from the login page to test. Then look up your pid (or just look in the url) and enter that for the survey.

### DIFF
--- a/dashboard-ui/src/components/AggregateResults/DataFunctions.js
+++ b/dashboard-ui/src/components/AggregateResults/DataFunctions.js
@@ -1562,7 +1562,6 @@ function getRatingsBySelectionStatus(data) {
                 break;
         }
     }
-    console.log(results);
     return results;
 }
 
--- a/dashboard-ui/src/components/App/index.jsx
+++ b/dashboard-ui/src/components/App/index.jsx
@@ -342,10 +342,14 @@ export class App extends React.Component {
                 }
                 else {
                     // generate a new pid by incrementing highest found
-                    const newPid = Math.max(...pLog.filter((x) =>
+                    const validEntries = pLog.filter((x) =>
                         !["202409113A", "202409113B"].includes(x['ParticipantID']) &&
                         x.ParticipantID >= LOW_PID && x.ParticipantID <= HIGH_PID
-                    ).map((x) => Number(x['ParticipantID']))) + 1;
+                    );
+                    // bug fix first one gets :LOW_PID, Math.max was returning negative infinty
+                    const newPid = validEntries.length > 0 
+                        ? Math.max(...validEntries.map((x) => Number(x['ParticipantID']))) + 1
+                        : LOW_PID;
                     const setNum = newPid % 24;
                     const participantData = {
                         ...SURVEY_SETS[setNum], "ParticipantID": newPid, "Type": isTester ? "Test" : "emailParticipant",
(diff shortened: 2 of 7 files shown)


kaitlyn-sharo (commented):
The online only survey uses 202411500 through 202411799 which shouldn't be a problem since these ids are 202501 prefixed. 

Just want to confirm no online-only participants will be in this collection?

kaitlyn-sharo (commented):
In terms of taking the text scenario and survey, everything seemed to work great and store great. In terms of results pages that automatically add this data, I believe the following changes should be made:

SurveyResults/resultsTable.jsx (delegation survey results):
const EVAL_MAP = {
    3: 'MRE',
    4: 'DRE',
    5: 'PH1',
    6: 'JAN25'
}

In TextBasedResults/TextBasedResultsPage.jsx, I was having some trouble accessing the charts. I kept getting null errors. I think either we have to exclude more keys (like kdmas and group targets), or this is sufficient to at least get rid of the errors (replace lines 258 and 259):
formatted[page['_id']][time_key] = Math.round(page[key]?.['timeSpentOnPage'] * 100) / 100;
obj[time_key] = Math.round(page[key]?.['timeSpentOnPage'] * 100) / 100;

Participant Progress table still looks good.

Nothing else has results yet because we don't have the eval 6 sim data, so no problems there.

brianpippin:
> The online only survey uses 202411500 through 202411799 which shouldn't be a problem since these ids are 202501 prefixed.
> 
> Just want to confirm no online-only participants will be in this collection?

Should only be participants with OSU, no online only data collect.

nextcen-dgemoets:
Some thoughts/questions:
- I assume we don't need a Delegation Survey version in the admin dashboard because it's the same survey?
  - Hmm, although after doing the delegation survey with my new PID and checking out the delegation survey results at `/survey-results`, the `Eval` dropdown contains `6 - undefined`.  This had my participant ID with a survey version of 5 in the table.
  - And I suspect my new pid is included in the Graph View for 5.x.  Is that what we want, or not?
- I'm getting a crash in `/text-based-results` when I click on the Participants tab:  `Cannot read properties of null (reading 'timeSpentOnPage')`
- Unrelated to this PR, and I'm not sure how I managed this, but I managed to select both medics in the delegation survey:
both-medics
  - I'm pretty sure this was just a UI thing and only one was actually selected / sent.

dereknop:
>   * Hmm, although after doing the delegation survey with my new PID and checking out the delegation survey results at `/survey-results`, the `Eval` dropdown contains `6 - undefined`.  This had my participant ID with a survey version of 5 in the table.

This should now be fixed

>   * And I suspect my new pid is included in the Graph View for 5.x.  Is that what we want, or not?

 Good question, I do not know the answer

dereknop:
> * I'm getting a crash in `/text-based-results` when I click on the Participants tab:  `Cannot read properties of null (reading 'timeSpentOnPage')`
>   * It's always this is because of something ugly in my local DB.  Can anybody else reproduce?

Couldn't reproduce, either this is the same bug Kaitlyn gave me a fix for or it's probably a local db issue

> * Unrelated to this PR, and I'm not sure how I managed this, but I managed to select both medics in the delegation survey:
>   both-medics
>   
>   * I'm pretty sure this was just a UI thing and only one was actually selected / sent.

I have no idea how this managed to happen but yes only one response will ever be saved by SurveyJS on radio type questions

kaitlyn-sharo:
> > * I'm getting a crash in `/text-based-results` when I click on the Participants tab:  `Cannot read properties of null (reading 'timeSpentOnPage')`
> >   
> >   * It's always this is because of something ugly in my local DB.  Can anybody else reproduce?
> 
> Couldn't reproduce, either this is the same bug Kaitlyn gave me a fix for or it's probably a local db issue
> 
Same bug I gave you a fix for.

nextcen-dgemoets (approved):
LGTM.  The identified issues (required for Monday's data collect) have all been addressed.

kaitlyn-sharo (approved):

PR #243, opened 2025-01-31
Itm 778 automated testing

kaitlyn-sharo:
This PR contains the setup for UI unit tests. Before you start running tests, rebuild your docker containers and make sure your frontend still works. 

There was a major server setup change in this PR. Before, we were importing our mongo database directly into our schema, but I've moved that out so that the schema can accept any database through the "context" parameter. Everything works the same way, essentially, but this allows us to inject a mock database into our same schema for better, stronger tests. Please test the UI rigorously (using your normal method of setup) to ensure that everything still works for you.

To run tests, check out the readme, which I've attempted to update with all the information necessary for basic testing.

Here is a summary of the tests included in this PR (for more details, look through the four test files for the suite and individual descriptions):

EmailEntry.test.jsx - Ensure that the /participantText page works as intended 
- There should be no option to sign in or create an account
- If the emails entered are not the same, an error should be given
- Email matching should be case insensitive
- A valid 'sign-in' (i.e. matching emails) should navigate to /text-based and generate a new PID. An email entry that has already been entered should generate the _same_ PID
- PID lookup should return the expected PID for the email

Render.test.jsx - Simply tests that the UI renders

Permissions.test.jsx - 
- test invalid login
- test new user gets sent to login page
- test 'return to login' button functionality
- test new account with duplicate email/username errors
- test that a user who is not logged in cannot access any routes except remote-text-survey
- test that a user who is logged in but not approved cannot access any routes except remote-text-survey and awaitingApproval
- test that a user who is approved, has logged in, but then logged out, cannot access any routes except remote-text-survey

Admin.test.jsx -
- tests that an administrator can access all available pages

You may notice for some of the routing tests that some routes are commented out. These routes do not work on prod, development, or in the tests and need to be investigated further. I spent some time on it but am making a ticket to complete it, as I was not able to make progress.

Found a bug while writing tests (yay!):
1. Accounts had been able to be created with duplicate emails and usernames. Now an error should appear. This should be case insensitive (and is tested in the unit tests)

Side note: When this gets merged, it should probably have all commits squashed, as there was a lot of trial-and-error during the process and things got pushed only so that we could work together.


dereknop:
In the readme shouldn't we run `npm run clean` before setting up the tests?

dereknop (approved):
105/105 passed for me. No more inconsistent behavior on which tests were passing and failing (tried several times). My first run had a bunch of fails but I believe that was because of what we discussed earlier where I didn't give it enough time after running `npm run setup-tests` before running `npm run test`. After giving it ~2 mins, everything works as expected

kaitlyn-sharo:
> In the readme shouldn't we run `npm run clean` before setting up the tests? Or are you saying it's just good practice to do that after testing to clean up the workspace?

It really only matters if you're having trouble running anything. For example, maybe you had locally installed dependencies on an older node version (i.e. 12) and switching node versions is causing problems with installs. It's a helpful troubleshooting tool, but I don't think it's a necessary step each time, especially since installing everything takes awhile.

dashboard-ui/jest.config.js
```diff
@@ -0,0 +1,27 @@
...
+        "/node_modules/(?!@adobe/css-tools)/"
+    ],
+    moduleNameMapper: {
+        '\\.css$': 'identity-obj-proxy', // Mock CSS imports with identity-obj-proxy
+        '\\.(xlsx|pdf|docx|png|R)$': '<rootDir>/src/__mocks__/fileMock.js', // Mock .xlsx, .pdf, .docx, .png, and .R files
+    },
+    preset: 'jest-puppeteer',
+    maxWorkers: 1,
+    globals: {
+        "provess.env": {
```
nextcen-dgemoets, line 22: Presumably this should be `process.env`?

kaitlyn-sharo:
closing this PR as 844 includes it and improves on it, fixing breaking bugs

PR #244, opened 2025-02-03
Adm probe responses download all fix

dereknop:
The download all button on http://localhost:3000/adm-probe-responses was not accurately reflecting the data in the individual tables (instead it was duplicating data and overwriting what should actually be there). The individual table downloads were already working. To test, download all and compare the excel sheet to what is shown in the UI.


kaitlyn-sharo (approved):
This seems to be working for me

nextcen-dgemoets (approved):
This seems to be working for me.  Before, from what I could see, the four ADMs had the same data, but in this branch, the data is different.

PR #245, opened 2025-02-03
Itm 845 permissions tests

kaitlyn-sharo:
I added some new tests to the UI. This includes improvements to authentication handling as well, allowing us to set up test users (with usernames, emails, passwords, and permissions) to test much easier. The new tests include:

- Non-admin approved users should be able to access only allowed routes
- Experimenters should be able to access only allowed routes
- Evaluators should be able to access only allowed routes
- ADEPT users should be able to access only allowed routes
- If a user is rejected, their awaitingApproval page should have different text
- ADEPT users and admins should see 3 extra columns in the progress table. No other user should see these colums.

While creating these tests, some bugs were found with routing and experimenter/ADEPT dashboard display. This is fixed in this PR. Look around the UI as an ADEPT user (it should be the same as an evaluator, plus the 3 extra columns in the progress table) and as an experimenter (should also be the same as an evaluator, plus pid-lookup).


dereknop:
On first pass, 191/191 passed. Haven't looked at code yet.

kaitlyn-sharo:
> I think this branch needs the cross-env fix from 778.

Yes, it will. I was waiting until all the windows-related fixes were done

dashboard-ui/src/__tests__/ApprovedRoutes.test.jsx
```diff
@@ -0,0 +1,184 @@
...
+            if (!res) {
+                if (isAdmin)
+                    await loginAdmin(page);
+                else if (isEvaluator)
+                    await loginEvaluator(page);
+                else if (isExperimenter)
+                    await loginExperimenter(page);
+                else if (isAdeptUser)
+                    await loginAdeptUser(page);
+                await testRouteRedirection(route, route, true);
```
dereknop, line 63: Am I understanding correctly that if it the route redirection doesn't work because of a sign out the first time we try again but if it doesn't work the second time _then_ the test fails?

dashboard-ui/src/__tests__/ApprovedRoutes.test.jsx
```diff
@@ -0,0 +1,184 @@
...
+describe('Route Redirection and Access Control Tests for evaluators', () => {
+    // log in as evaluator
+    beforeAll(async () => {
+        await loginEvaluator(page);
+    });
+
+    runAllowedRoutesTests(false, true);
+    it('Evaluators should not see extra headers on progress table', async () => {
+        await page.goto(`${process.env.REACT_APP_TEST_URL}/participant-progress-table`);
+        await page.waitForSelector('text/This research was developed');
```
dereknop, line 111: I think we could have some of these repeated strings as constant variables

dashboard-ui/src/__mocks__/testUtils.js
```diff
@@ -88,4 +88,34 @@ export async function testRouteRedirection(route, expectedRedirect = '/login', r
     }
```
dereknop, line 88: Is the screenshot on line 27 necessary? I saw the comment above it but I just don't understand haha

dereknop (approved):

dereknop:
Everything worked and I don't see any issues, was just confused about a few things.

dashboard-ui/src/__mocks__/testUtils.js
kaitlyn-sharo: It is necessary. I thought that was a 778 change, but maybe I'm just getting all confused now. It needs some more concrete browser connection, from all of my experimentation, and the screenshot is a fast, efficient way to get that

dashboard-ui/src/__tests__/ApprovedRoutes.test.jsx
kaitlyn-sharo: That's a good idea. I'm working on a refactor of index.jsx now, so I'll just include that since it falls under the "refactor" category

dashboard-ui/src/__tests__/ApprovedRoutes.test.jsx
kaitlyn-sharo: Yes. This is what was done in 778 initially because it seems that the browser kicks us out periodically, resetting our user and storage, etc., so in the middle of a test, we are just randomly logged out. If we get logged out, we just log back in and try again. If it fails again, then it's a bad test. This is what fixed those random errors we were getting. If you noticed, even though different tests failed each time, it was usually right around the same time-point, as if jest gives us 15 seconds and then says "hey let's restart everything!". There may be a configuration option somewhere to prevent that, but I have not found it.

kaitlyn-sharo:
This should now be updated with the current windows-related fixes

kaitlyn-sharo:
Note to other reviewers: The retries in the tests and screenshot are removed in my next branch (with the refactor).

kaitlyn-sharo:
closing this PR as 844 includes it and improves on it, fixing breaking bugs

PR #248, opened 2025-02-11
Itm 852 jan eval

kaitlyn-sharo:
https://github.com/NextCenturyCorporation/itm-ingest/pull/106

Run the ingest PR first. Then run this. Check that all pages are still working. 
/survey-results
/text-based-results
/humanSimParticpant - nobody took qol sim
/humanProbeData
/human-results

Check the RQ pages. We want RQ 1/3/4, 5, 6, and 8 to all have options to view DRE and Jan Eval data now, but only if Phase 1 is chosen. All data should appear. Note that not all participants completed the survey, so some data will _appear_ to be missing. No columns should be completely empty (except QOL Sim KDMAs in /humanSimParticipant). Some cells will be missing data if the participant did not complete the survey or did not complete a particular sim.

### DIFF
--- a/dashboard-ui/src/components/AggregateResults/DataFunctions.js
+++ b/dashboard-ui/src/components/AggregateResults/DataFunctions.js
@@ -144,6 +144,7 @@ const ATTRIBUTE_MAP = {
 const TEXT_MEDIAN_ALIGNMENT_VALUES = {};
 const SIM_MEDIAN_ALIGNMENT_VALUES = {};
 const SIM_ORDER = {};
+export const POST_MRE_EVALS = [4, 5, 6];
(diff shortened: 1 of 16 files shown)


dereknop (commented):
Survey results works (only 7xx results being shown for Jan Eval toggle), ditto for text-based results. `/humanProbeData` and `humanProbeData` look as expected. RQ tables all look good to me as well.

dereknop (approved):

nextcen-dgemoets (approved):
LGTM.  I think everything's there that's supposed to be, but I'm certainly not 100% on what rows/columns/cells are expected to be empty.

The only thing that was a tiny bit surprising was that  in `/human-results`, when I selected `vol-ph1-eval-4`, there were no PIDs associated with it.  The other VOLs had PIDs (although only one or two) and the QOLs, as expected, had none.

brianpippin (approved):

PR #249, opened 2025-02-14
Dev to Main

brianpippin:


dereknop:
Added node_modules to the .gitignore since people will have that locally from the tests. Fixed the issue with some of Adept's text being cut off early in survey

dereknop (approved):

nextcen-dgemoets (commented):
I'm still not sure why two participants have 10 and 8 text scenarios completed.  But maybe it's not a problem?

Also, can there be PIDs that start with 2025 but aren't in the Jan 2025 Eval Data dataset?  It seems so?

nextcen-dgemoets (approved):
Other than the 10 and 8 text scenarios completed thing, looks good to me.  If there are actual data problems, I'm not likely to spot them.

kaitlyn-sharo (approved):

PR #250, opened 2025-02-18
add dre data to 134 for jan eval

kaitlyn-sharo:
Run the ingest PR (https://github.com/NextCenturyCorporation/itm-ingest/pull/108).

Look at RQ 1/3/4 to make sure the DRE columns populate for Jan eval (using both the toggle and the dropdown)

This also removes rows with an empty last column from RQ6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ docker_setup/ServerCertificate.crt
 docker_setup/darpaitm.key
 .DS_Store
 node_modules
+package-lock.json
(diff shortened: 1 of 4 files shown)


dereknop (approved):

nextcen-dgemoets (approved):
LGTM.

RQ6 rows no longer have an empty right-most row.
RQ1/3/4 have DRE data for the non-comparison ADM Types in the Jan 2025 Eval (whether by toggle or dropdown), but they didn't on `main`.

I did notice that in RQ1/3/4, the ADM Loading columns say "exemption" but the variable definitions call them "exceptions".

PR #251, opened 2025-02-19
added delegation-by-block table to exploratory analysis

kaitlyn-sharo:
Check out the exploratory analysis section. At the bottom, there is this new table, called "Delegation Data By Block". 

The instructions were to copy the delegation results table and add a few columns:
1. Alignment of Delegator|Observed_ADM using DRE and PH1 server for each Block/DM (two columns per block/dm)
2. Alignment difference of Delegator|Observed_ADM using DRE and PH1 for each comparison section

The results table should not have changed.

The table in the exploratory analysis section should have all the new columns and follow the dropdown and checkboxes, just as the other tables do.

Variables should be updated to match.


dereknop (approved):

nextcen-dgemoets (approved):
LGTM.  I will observe that according to Chrome, my /exploratory-analysis tab is taking ~1.6 GB.  Sometimes the tab reacts sluggishly.  On prod, it takes 207 MB.  It's not completely apples-to-apples because there may be more differences since prod was last updated, but still...it's a lot of data.

kaitlyn-sharo:
> LGTM. I will observe that according to Chrome, my /exploratory-analysis tab is taking ~1.6 GB. Sometimes the tab reacts sluggishly. On prod, it takes 207 MB. It's not completely apples-to-apples because there may be more differences since prod was last updated, but still...it's a lot of data.

My last two tables also take forever to load. I found it strange because the delegation survey table by itself doesn't seem to take a long time...open to suggestions!

PR #252, opened 2025-02-19
Dev to Main

brianpippin:


dashboard-ui/src/__tests__/ServerEndpoints.test.jsx
```diff
@@ -0,0 +1,394 @@
...
+    }, 10000);
+
+    it('should create a new SoarTech session successfully', async () => {
+      const response = await axios.post(
+        `${SOARTECH_URL}/api/v1/new_session?user_id=default_user`
+      );
+
+      expect(response.status).toBe(201);
+      expect(response.data).toBeTruthy();
+    }, 10000);
```
kaitlyn-sharo, line 50: @dereknop do you want to change these ST server timeouts before we merge to main?

kaitlyn-sharo:
Changes include:
- Adding unit tests
- Refactoring the graphql code to inject the dashboard rather than using a static import
- Refactoring index.jsx (in App) to be function-based instead of class-based, getting rid of console errors during reroutes and fixing all other routing issues (for example, /random-link should redirect to the user's 'home-base' page, but was not before)
-  A few minor bug fixes based on bugs found during the creation of unit tests and during other testing (qqol -> qol typo fix, a few components that would error on receipt of incomplete data instead of gracefully failing, duplicate usernames/emails were allowed when creating accounts if capitalization was different)
- Remove useless RQ6 rows

dashboard-ui/src/__tests__/ServerEndpoints.test.jsx
dereknop: Yea I'll up them to 40

kaitlyn-sharo (approved):

dereknop (approved):

nextcen-dgemoets (commented):
Overall, LGTM.  I re-ran the tests and got one failure (timeout), but it was while playing with the Exploratory Analysis page that at one point made my browser freeze, so that probably wasn't a fair test.

I re-ran after closing the tab, and got the same failure.

Note that I *didn't* do a clean/set-up prior to this run, but I don't think anything changed here since yesterday that would require that?

nextcen-dgemoets (approved):
I did a clean / setup / re-run, and all of the tests passed.

PR #253, opened 2025-02-19
Delegator_mil column Jan Eval Updated

dereknop:
"We need to create an override in the UI (the RQ Tables), that says if a column is describing the answer to Military or Civilian and the data is coming from Eval 6, then we override their answer to always be Military.  We do not want to change the database.  We want to persist this into the download features as well."

No changes to database. The `Delegator_Mil` column in the research tables should be `yes` for all Jan 6 Eval participants. This includes when phase 1 eval is selected but `Include Jan 2025 Eval Data` is checked. Also download the data to make sure this change persists there.


kaitlyn-sharo (approved):

nextcen-dgemoets (approved):
LGTM.  I went to the various tables, selected "Include Jan 2025 Eval Data," filtered on PID 2025, and filtered on `Delegator_mil` of no, and in all cases, got zero rows.

PR #256, opened 2025-02-21
Adept Qualtrix tests

kaitlyn-sharo:
This does some basic testing on the adept qualtrix page and text-scenarios page. 

- checks that no matter what key is pressed, the site does not quit/return to the beginning (in adept qualtrix text, survey, and email-based text)
- checks that adept qualtrix goes from text scenario into the survey
- checks that the text scenario and survey can be completed through the adept qualtrix method
- mocks text scenario and survey data (not really a mock as much as pulling data from the database and shortening it to make testing faster) - this sets us up for many more possible tests in the future

Note that one of the tests here is incredibly long because it ends up sending all of the alignment data to both servers. It should have enough of a timeout not to error, but you might be sitting for a few minutes waiting for that one.

### DIFF
--- a/dashboard-ui/src/__mocks__/mockData.js
+++ b/dashboard-ui/src/__mocks__/mockData.js
@@ -1,4 +1,9404 @@
-export const textConfigMock = {
+// includes 3 default adms, shortened scene actions and patients
+export const surveyV5 = {
+    _id: 'delegation_v5.0',
+    survey: {
+        pages: [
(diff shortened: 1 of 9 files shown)


nextcen-dgemoets (approved):

dereknop:
All tests passed on my first try. Looking through code before approving

dereknop (approved):

PR #260, opened 2025-03-04
Remove empty chart from /results

kaitlyn-sharo:
In pre-DRE evals, the RQ2 table was showing up empty on the /results page. Those empty tables are now replaced with simple text about what the user needs to do in order to begin seeing results.


dereknop (approved):
Makes sense, looks good

nextcen-dgemoets (approved):

PR #261, opened 2025-03-05
Itm 851 windows cleanup

kaitlyn-sharo:
Does not need to go into dev to main. This is more of a backlog-maintenance type ticket.

Prior to this, windows users would have to manually clean their dashboard directories of node_modules and package-locks. This updates the `npm run clean-win` script so that it actually works on windows. Ports should now clear and directories should be cleaned when this is run. 

If you have not run a setup-tests (in which case, your node_modules will not have the required npm dependencies), you may be asked to install kill-port and rimraf during the running of the script.

To test, on a Windows laptop, run `npm run setup-tests` and `npm run clean-win`. You should end up with no node_modules or package-locks.


dereknop (approved):
`package-lock.json` and `node_modules` generated during setup-tests were removed during `npm run clean-win`

kaitlyn-sharo:
Will wait to merge until dev-to-main is completed

PR #265, opened 2025-03-20
Itm 908

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?selectedIssue=ITM-908)

To test, first, you'll need to rebuild the gql container. Then go to http://localhost:3000/admin and make sure the lists of users are still being populated (This ensures the query is still working for admins). I have added an automatic query to `index.jsx` so that we can easily test that this query no longer works for non-admins. This will all be deleted before merging. 
To test non-admin account:
- Create a new account and then approve it (but do not give it any privileges) using your pre-existing admin account
- Change the hard-coded password on line 183 of `index.jsx` to the password of your new account
- Sign back in with the new account that is now approved. You should see this in the console:Screenshot 2025-03-20 at 2 11 20 PM
- At this point you can double check it works for admin accounts by signing out and changing the hard coded password to your admin account password and then signing back in: Screenshot 2025-03-20 at 2 18 38 PM

Again, all of the changes to `index.jsx` will be deleted prior to merging. I just thought it would make it easier to test since I couldn't get the sandbox working on 9100.


nextcen-dgemoets (approved):
LGTM!  I followed the steps exactly and got the expected results:
itm-908

kaitlyn-sharo (approved):
Worked as described for me

PR #266, opened 2025-03-20
Itm 861 rq5 update

kaitlyn-sharo:
[See the ticket](https://nextcentury.atlassian.net/browse/ITM-861?atlOrigin=eyJpIjoiZGY2MDA5MjZmYjU0NGU2MzlhMTgwZGFiMjY3MzI0YWQiLCJwIjoiaiJ9)

This task had been completed in ingest originally. All the data was reviewed and given to Jennifer. We had not wanted to put it in the dashboard on prod due to long load times, but that decision has been reversed due to demand for data.

All of this data exists in the database already due to the previous ingest script, so you should just be able to load this up and see the data in the dashboard, rather than in csv form (which we had originally decided to go with instead of adding this table).

The table only exists for PH1. I took away the Jan Eval and DRE checkboxes because their rows were practically empty. Jennifer does not want their data here. Jan Eval and DRE have the original RQ5 table. All variable definitions should match.


dereknop (approved):
Looks good, like Darren mentioned it takes a bit to load but I'm not sure what we can do about that given how much data it is 
Screenshot 2025-03-24 at 9 00 24 AM

PR #268, opened 2025-03-24
Survey Results Endpoint Security Update

dereknop:
A `user` field was attached to the survey results documents (not always) that contained username and email information. I changed the endpoints that pull the survey results to omit this field. I added a console log to http://localhost:3000/survey-results that shows the data we are pulling to the front end. Glance through it real quick to make sure you don't see a `user` field in any of that data. I will delete this console log before merging. You will need to rebuild gql container.


nextcen-dgemoets (approved):
I did some spot checking of the output, and didn't see anything personally identifying like a username.

I did get the following error, which is probably known:
page-not-found

kaitlyn-sharo (approved):

PR #269, opened 2025-03-25
KITWARE EXPERIMENT: Itm 894 p2e research

kaitlyn-sharo:
Run https://github.com/NextCenturyCorporation/itm-ingest/pull/118 first

Check out RQ2. There should be a new drop down for Phase 2. The other phases should remain unchanged. Phase II should only show RQ 2.3 with a new format. 

Will need to update when all adm names are set so I can separate baseline from aligned.


dereknop:
Phase II Experiment 1 showing up on rq2 page and the table is populated after running ingest. One small suggestion I would make is to capitalize the first letter of the values in the `ADM Type` and `Target Type` columns.

dereknop (approved):

nextcen-dgemoets (approved):
LGTM.  I probably wouldn't recognize if there was an error in the data, but the table itself looks good and I didn't see any problems.

PR #270, opened 2025-03-26
Itm 915

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-915)

Run [ingest](https://github.com/NextCenturyCorporation/itm-ingest/pull/119) first 

http://localhost:3000/results
http://localhost:3000/adm-results

Should all be functional now. Multi-kdma should not be an option for /adm-results. You'll want to grab the newest backup of the db and also need to rebuild gql container.


nextcen-dgemoets (commented):
This all looks great.

The only thing that made me pause is how the "synthetic" alignment target ids (e.g., `target1` through `target498`, the rows of the csv) are in the dashboard.  I just generated those names because they needed to be unique, and traced back to the csv file, but they're not very informative in the dashboard.  I wonder if there's a way to use the PID from the csv to identify from which human that synthetic alignment target is derived?

In retrospect, maybe I could have named these alignment targets `[pid]-[eval]-[type]`, e.g., `target1` would become `20249201-4-overall`.  That would require updating the experimental server branch that Kitware is using right now...maybe it's too late.

dashboard-ui/src/components/Results/resultsTable.jsx
```diff
@@ -348,7 +348,7 @@ class ResultsTable extends React.Component {
                                 }
                             </Query> :
                             <>
-                                {this.state.evalNumber >= 4 ?
+                                {(this.state.evalNumber >= 4) && (this.state.evalNumber != 7) ?
```
kaitlyn-sharo, line 351: This should show the new table from my PR if evalNumber==7

kaitlyn-sharo (commented):
I think sorting the targets differently (target1, target2, target3 rather than target 1, target 10, target 100) would be really nice. However this did take away eval 7 where it wasn't supposed to be and made the data visible on the other two pages. If you think the target sorting is a different ticket, happy to approve.

dereknop:
> I think sorting the targets differently (target1, target2, target3 rather than target 1, target 10, target 100) would be really nice. 

Frustrating that the built in sort function doesn't do this on its own. Should be good now.

dashboard-ui/src/components/AdmCharts/scoreChart.jsx
```diff
@@ -68,15 +68,17 @@ class ScoreChart extends React.Component {
                     if (error) return <div>Error</div>
 
                     const chartData = data[GET_HISTORIES_BY_ID]?.filter((x) => x.evalNumber == this.props.evalNumber);
+                    
 
-                    let admNameParamter = "ADM Name";
+                    console.log(chartData)
```
kaitlyn-sharo, line 73: remove console log

kaitlyn-sharo (approved):
Target numbers are organized prettier now! I think everything works as intended

nextcen-dgemoets:
> The only thing that made me pause is how the "synthetic" alignment target ids (e.g., `target1` through `target498`, the rows of the csv) are in the dashboard. I just generated those names because they needed to be unique, and traced back to the csv file, but they're not very informative in the dashboard. I wonder if there's a way to use the PID from the csv to identify from which human that synthetic alignment target is derived?
> 
> In retrospect, maybe I could have named these alignment targets `[pid]-[eval]-[type]`, e.g., `target1` would become `20249201-4-overall`. That would require updating the experimental server branch that Kitware is using right now...maybe it's too late.

@brianpippin  Do you or Jennifer want to comment on the target IDs appearing as `target1` through `target498` in the dashboard?  Is there anything we should do about that, either for this ticket or a subsequent one?  If not, then this ticket is good to go.

brianpippin:
> > The only thing that made me pause is how the "synthetic" alignment target ids (e.g., `target1` through `target498`, the rows of the csv) are in the dashboard. I just generated those names because they needed to be unique, and traced back to the csv file, but they're not very informative in the dashboard. I wonder if there's a way to use the PID from the csv to identify from which human that synthetic alignment target is derived?
> > In retrospect, maybe I could have named these alignment targets `[pid]-[eval]-[type]`, e.g., `target1` would become `20249201-4-overall`. That would require updating the experimental server branch that Kitware is using right now...maybe it's too late.
> 
> @brianpippin Do you or Jennifer want to comment on the target IDs appearing as `target1` through `target498` in the dashboard? Is there anything we should do about that, either for this ticket or a subsequent one? If not, then this ticket is good to go.

Is there anything that maps "targetN" to the values used for that target?

nextcen-dgemoets:
> Is there anything that maps "targetN" to the values used for that target?

It's the order of the rows in the csv file, which appears to be the order the ADEPT scenarios (at least those with all values filled in) are returned from the `userScenarioResults` table.  Perhaps a one-off script can update the alignment target IDs in those tables based on this?  I can't imagine we want to update the server branch to set the target IDs this way-- unless of course we envision Kitware wanting to re-run things.

brianpippin:
> > Is there anything that maps "targetN" to the values used for that target?
> 
> It's the order of the rows in the csv file, which appears to be the order the ADEPT scenarios (at least those with all values filled in) are returned from the `userScenarioResults` table. Perhaps a one-off script can update the alignment target IDs in those tables based on this? I can't imagine we want to update the server branch to set the target IDs this way-- unless of course we envision Kitware wanting to re-run things.

I am asking because if ADEPT or KITWARE look at the table I want it to be easy to see what is going on and what the values are, even if there is a key for the values or something.

dashboard-ui/src/components/Results/resultsTable.jsx
kaitlyn-sharo: This can be done now

nextcen-dgemoets:
> > > Is there anything that maps "targetN" to the values used for that target?
> > 
> > It's the order of the rows in the csv file, which appears to be the order the ADEPT scenarios (at least those with all values filled in) are returned from the `userScenarioResults` table. Perhaps a one-off script can update the alignment target IDs in those tables based on this? I can't imagine we want to update the server branch to set the target IDs this way-- unless of course we envision Kitware wanting to re-run things.
> 
> I am asking because if ADEPT or KITWARE look at the table I want it to be easy to see what is going on and what the values are, even if there is a key for the values or something.

I really don't know if this will be sufficient for Adept and Kitware.  I feel like maybe they'll need a little more, but I'm not sure exactly what.  I guess we can just hand it over to them and let them ask for something, but I was trying to get ahead of it if there was something we knew made sense to give them.  Or maybe when Jennifer looks at all of this next week, she'll ask for something to supplement what we have now in this table or the RQ2 table.

nextcen-dgemoets (approved):
LGTM.  As far as I know, this is everything we need here.  Maybe when Jennifer looks at the data next week she'll have insight as to whether KW and AD will be able to look at the table and have it be easy to see what is going on and what the values are.

It strikes me that in the ADM pages, it's called "Multi-kdma Experiment" but in RQ2, it's called "Phase II Experiment 1".  But that's not a function of this ticket.

PR #272, opened 2025-03-31
Itm 911

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-911)

I tracked down the following github issues/documentation to get this library to work with our version of Apollo. So the implementation doesn't exactly match the simple example in the README on the library's github page. 
[implementing with apollo v3](https://github.com/slicknode/graphql-query-complexity/issues/81)
[using requestDidStart hook](https://github.com/MichalLytek/type-graphql/blob/4501867fffe3e6f5b3e71af0b71651efcd48d9c3/examples/query-complexity/index.ts#L16-L64)

I gave all of the queries complexity values that mirrored the amount of data they are actually retrieving (these are easy to change/add for new queries). 

I set the maximum query complexity to `1000`. The largest query we currently use, `GetConfigs` comes out to `550` in complexity based on the complexity values of the queries included. This retrieves the images and large json files for the survey and text based scenarios. 

To test, rebuild your gql container. Make sure by clicking around that queries are still loading properly and data is being displayed as it normally would. Then to make sure the blocking functionality is working I would increase the complexity of a query in `server.schema.js`, rebuild the gql container, and then check the docker logs to see the logs confirming the request was blocked due to the complexity being too high. For example, you could set `getAllSurveyConfigs`, `getAllTextBasedConfigs`, and `getAllTextBasedImages` to a complexity of `500` each. Then you should see the following message in the docker logs: `Operation "GetConfigs" blocked: complexity of 1500 exceeds threshold: 1000`

I also took this opportunity to clean up the `server.schema.js` file by removing a bunch of endpoints from MVP that we have never used (most of which pointed to collections that don't exist).


dereknop:
Leaving as a draft as I incorporate [Itm-912](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-912)

nextcen-dgemoets:
> [rate limit from scratch approach](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/272/commits/812e07dcfdcb7c61812c5cc1ec386c14e9eb4db6)

Does this mean this ticket is ready to review / no longer Draft?

dereknop:
I have attached a Python script to test the rate limiting. From inside `/node-graphql` you can run `python tester.yp`. It should try to make 500 requests in quick succession and log the results. I have the limit set to 300 requests per limit (you can mess with this inside `rateLimiter.js`). 

Another way to test could be simply spam refreshing the home page, although you have to click fast to reach the limit within the time window. 

I will remove the test script before this gets merged. As a reminder, you'll need to rebuild the gql container if you mess with any of the limits while testing.

kaitlyn-sharo (approved):
I'll admit I don't really understand the code here, but nothing looks crazy and the tests don't break, which is great!

Python script ran as expected. 

Changing the complexity gives a 400 request error in the dashboard console and a more detailed error in docker. The network tab also shows a nice detailed error. 

Clicked around and didn't see any problems!

dereknop:
> How did you get/calculate query complexity? Maybe "based on the complexity values of the queries included" will be obvious to anybody maintaining this or adding a query. Just making sure we don't need to document how developers should assign complexity in future queries.

At first, I wanted the complexity to be a 1:1 match to the size of the data being returned by the query, but when I started going down that route, I realized it defeated the whole purpose because we can't look at the size of the data until the request is actually returned. So, instead, I manually looked at the size of the requests in the browser for each of our queries and gave them a reasonable weight. 

> Don't forget to delete the rate limiter log and the test script (though keep it somewhere for posterity-- maybe in `itm-extras`?)

Good call, I'll add it to the repo

PR #273, opened 2025-04-01
added pid to rq23 phase 2 table

kaitlyn-sharo:
Run https://github.com/NextCenturyCorporation/itm-ingest/pull/121

Check out the RQ 2 -> Phase II Experiment I -> RQ 2.3 table. PID has been added. The alignment variables have been renamed according to Jennifer's specifications. The variable download has been updated accordingly. The ADM Name filter has been enlarged so that the full name is visible.


nextcen-dgemoets (approved):
The data looks good-- having the PID really helps.

> The ADM Name filter has been enlarged so that the full name is visible

I just noticed that this AM but didn't say anything...and just like that, it's fixed!

dereknop (approved):
Looks good, same rows as Darren reported

PR #275, opened 2025-04-02
Itm 929

dereknop:
Run [ingest](https://github.com/NextCenturyCorporation/itm-ingest/pull/122) first. Rebuild the gql container. 

Small update to rq2.3 table to account for the change in adm names. I updated a few gql queries to use the top level data instead of digging through the history array of the documents. 

Make sure these pages still work as expected:
http://localhost:3000/results
http://localhost:3000/research-results/rq2 (phase 2)


nextcen-dgemoets (approved):
Looks a lot better!

kaitlyn-sharo (approved):
Worked for me

PR #278, opened 2025-04-07
mj and io targets in probe response table

dereknop:

The table should have two new columns for eval 7. Next to the `Alignment Target` column, there should be the corresponding kdma values that make up `targetx` in `MJ KDMA Target` and `IO KDMA Target`


kaitlyn-sharo (approved):
Code looks good, table looks good!

nextcen-dgemoets (approved):
LGTM.  The new columns are present and accounted for.  The column names in the dashboard (`MJ KDMA Target` and `IO KDMA Target`) are slightly different from the download (`MJ KDMA` and `IO KDMA`), but it's clear.

PR #279, opened 2025-04-07
Itm 946 baseline multikdma

kaitlyn-sharo:
Run https://github.com/NextCenturyCorporation/itm-ingest/pull/125 first

Check out the Phase 2 RQ2 table. Notice that ADM Type is gone, because we only have one row for every aligned ADM. Baseline ADMs are not given their own rows.

You should see nearly double the number of columns compared to development. Now for every MJ/IOx_KDMA/Alignment column, there are two - one aligned, one baseline. The aligned alignment columns should now be updated with PH1 server results (using the distance-based endpoint). Everything else for Aligned data should remain the same. 

The baseline columns should have exactly the same KDMAs and averages all the way down the first 8 columns. The last 4 columns should be different, as it's showing the baseline compared to each human's set of probes (overall, train, or narr).

Variable definitions were updated


nextcen-dgemoets (approved):
LGTM.  The data was as you explained.

The only tiny issue I saw was that `ADM Type` was still in the variable definitions.

kaitlyn-sharo:
> LGTM. The data was as you explained.
> 
> The only tiny issue I saw was that `ADM Type` was still in the variable definitions.

Updated, thanks for pointing that out

dereknop (approved):
LGTM

PR #282, opened 2025-04-11
Truncation column fix

dereknop:
http://localhost:3000/research-results/rq1

The `Truncation Error` column should only be visible if Jan Eval results are present. Try downloading the data when just phase 1 data is present, there should be no `Truncation Error` column. If you hit `Include Jan 2025 Eval Data` or swap the eval to Jan 2025 Evaluation and then download, the column should be present. Be sure to check the download and the table on the dashboard.


nextcen-dgemoets (approved):
LGTM.  It was as described in the PR description.
Was there a ticket for this?

brianpippin (approved):

PR #283, opened 2025-04-11
add rq2 ph2 filter for only rq134 pids

kaitlyn-sharo:
Jennifer requested a filter on the PH2 Exp2 RQ2 table that would only show the PIDs that appear in the RQ134 table. I used the logic from the `getRQ134Data` function in order to filter these PIDs. The checkbox is where the "Show DRE/Jan Eval Data" checkboxes usually are.


nextcen-dgemoets (approved):
LGTM.  The filter was there and affected the table and download.

dereknop (approved):

PR #284, opened 2025-04-11
removed extra columns from text based results tables

kaitlyn-sharo:

In the participant table tab, ST scenarios had two extra columns:
image
This ticket removes them.


nextcen-dgemoets (approved):
LGTM.  The spurious columns are gone from the table and the download.

dereknop (approved):
I see the extra columns on dev but not on your branch. LGTM.

PR #286, opened 2025-04-16
Itm 964

dereknop:
Small change to `util.js` to account for the changes to DRE scores. We are pulling their fixed distance-based scores from `mostLeastAligned`, which was replaced in [ingest](https://github.com/NextCenturyCorporation/itm-ingest/pull/129).

http://localhost:3000/research-results/rq1

Check that this column `DRE/Distance Alignment score (Delegator|target)` is populated when the `Include DRE Data` box is checked. Also, select `Dry Run Eval` from the drop-down and make sure this column `Alignment score (Delegator|target)` is populated.


kaitlyn-sharo (approved):
I see data appearing where I expect it!

brianpippin (approved):
LGTM

PR #291, opened 2025-04-28
ADM Probe Responses Table

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?jql=assignee%20%3D%2060ba425da547eb00686ee0ce&selectedIssue=ITM-976)

Look at MJ5 under Phase 1 and make sure that you can see multiple probe responses for Probe 3. Not every ADM sends more than one response, but most do (cross-reference with the `admTargetRuns` collection in Mongo Compass)


nextcen-dgemoets (approved):
On `development`, only one probe response was shown.  On this branch, you can see all probe responses (including in the download).

PR #293, opened 2025-05-15
Itm 1003

dereknop:
[Ticket](https://nextcentury.atlassian.net/browse/ITM-1003?atlOrigin=eyJpIjoiY2QwMmE2MDIzZWFiNDUyN2FmYzlkM2UwYzg2NzJmYjkiLCJwIjoiaiJ9)

Fixes the time columns for `Delegation Data by Block` and the Delegation Results tables. The format should now be in minutes. If someone's total time was 54 minutes and 45 seconds, we will mark that as `54.75`. I changed the variable documents to reflect this change. 

Check on 
http://localhost:3000/research-results/exploratory-analysis

and


dashboard-ui/src/components/SurveyResults/resultsTable.jsx
```diff
@@ -69,19 +70,13 @@ const STARTING_HEADERS = [
...
-function formatTime(seconds, includeHours = false) {
-    seconds = Math.round(seconds);
-    let hours = (Math.floor(seconds / 60 / 60) % 60);
-    hours = hours.toString().length < 2 ? '0' + hours.toString() : hours.toString();
-    let minutes = Math.floor(seconds / 60);
-    minutes = includeHours ? minutes % 60 : minutes;
-    minutes = minutes.toString().length < 2 ? '0' + minutes.toString() : minutes.toString();
-    let formatted_seconds = seconds % 60;
-    formatted_seconds = formatted_seconds.toString().length < 2 ? '0' + formatted_seconds.toString() : formatted_seconds.toString();
-    return includeHours ? `${hours}:${minutes}:${formatted_seconds}` : `${minutes}:${formatted_seconds}`;
```
nextcen-dgemoets, line 84: I feel like we shouldn't just delete the old code entirely.  Could we have a `formatTimeInMinutes` and `formatTimeInMMSS` or something like that?

dashboard-ui/src/components/SurveyResults/resultsTable.jsx
```diff
@@ -229,7 +227,7 @@ export function ResultsTable({ data, pLog, exploratory = false, comparisonData =
             const timeDifSeconds = (new Date(entry.timeComplete).getTime() - new Date(entry.startTime).getTime()) / 1000;
             obj['Total Time'] = entry.startTime ? formatTime(timeDifSeconds, true) : null;
```
nextcen-dgemoets, line 228: Could this be `Total Time (Minutes)`?

dashboard-ui/src/components/SurveyResults/resultsTable.jsx
```diff
@@ -44,7 +44,8 @@ const STARTING_HEADERS = [
     "End Time",
     "Total Time",
```
nextcen-dgemoets, line 45: Could this be `Total Time (Minutes)`

nextcen-dgemoets (commented):
Well, I think from a human factors / UI standpoint, this is objectively worse.  From an ease of analyzing the downloaded file, this is objectively better.  There's probably a solution somewhere that involves leaving the dashboard alone but using some Excel macro to convert MM:SS to float minutes...but obviously we need to go with this solution for expediency.

The only thing I would ask is that we add `(Minutes)` in some places it isn't already, like the Total Time in the survey results and Delegation Data by Block table.  And I don't think we should delete the old code entirely.

brianpippin:
I agree with Darren about this being a new column next to the old column instead of replacing it, perhaps that is too many columns.  A future ticket, could be to toggle the visible column from time elapsed total to time elapsed minutes or something.

dereknop:
Both formats now included

nextcen-dgemoets (approved):
LGTM!

PR #295, opened 2025-05-30
Itm 1006

dereknop:
NOTE: Similar to Darren's server pr, you must upload the ADEPT YAML files to their server locally and edit your `.env` file to point to the local server accordingly 

Run the [ingest pr](https://github.com/NextCenturyCorporation/itm-ingest/pull/136) first.

I would first check that the review page is showing the new phase 2 scenarios http://localhost:3000/review-text-based

Then go to the login page to access the text scenarios:
Screenshot 2025-05-30 at 9 45 57 AM

There are a few things that I will delete before merging that are added in to make for easier testing. For example, below you will see `June2025-AF2-eval` to see what scenario is being presented to you as well as the probe id. Both of these will be removed before merging. 
Screenshot 2025-05-30 at 9 47 09 AM

I also have some console logs that demonstrate the order of the probes for each scenario being randomized:
Screenshot 2025-05-30 at 9 49 15 AM

Once you click through the scenarios, you should see your results uploaded to `userScenarioResults`:
Screenshot 2025-05-30 at 9 54 33 AM

Lastly, go to http://localhost:3000/pid-lookup and enter the email you used to complete the text-based scenarios. Verify that your PID shows up. OSU will need this to enter the participant VR runs.

Screenshot 2025-05-30 at 9 56 53 AM


dereknop:
> What is `template Probe N` in the `questions` under `Probe N` of each object? It looks empty. Related, What is `probe` under `Probe N` -> `questions` -> `probe Probe N` (and `template Probe N`? Also looks empty.

`template Probe N` is just the SurveyJS element that renders the probe info. SurveyJS treats it like any other question, even though it isn't really a question (which is why it is attached to the document that gets uploaded). I could parse those out before the document gets uploaded, I just hadn't bothered since it hadn't impacted anything negatively. The `probe: ""` field you mention used to be used, but is no longer necessary, so I can remove that since it is just noise now.

dereknop:
> I tried repeating the survey with the same email address / PID. It created 4 new entries in the DB, including a second MF2 because I believe it was randomly selected both times. Is this the expected behavior?

Did it prompt you that the email had already been used? The behavior I would expect is this pop-up:
Screenshot 2025-06-01 at 7 30 52 PM

nextcen-dgemoets:
> Did it prompt you that the email had already been used? The behavior I would expect is this pop-up:
> <deleted>

Yes it did.  Maybe this is all we need.

nextcen-dgemoets (approved):

PR #296, opened 2025-06-02
Dev to Main

brianpippin:
* june data collec baseline

* split text for injury and role. Make text flush across cells. scenario description as prop

* supplies first draft

* comparison test

* comparison better approach

* .

* text scenario progress

* test

* review page

* sorted

* phase2text styling

* clear

* not needed

* ph1

* new login handler, use new scenarios

* pid lookup update

* posting to ta1 server

* mostLeastALigned, KDMA scores

* randomize probe order

* fixed 8 documents uploaded bug

* un edit .env

* probe field removed, hyphens

* stuff to remove before merging


dereknop (approved):
http://localhost:3000/review-text-based looks as expected.

I went through the experiment a few times based on Jennifer's document for OSU and did not run into any issues taking the scenarios, verifying the scores were attaching to the mongo documents, and then looking up the PID using the email. 

The only issue I had was needing to delete my adept server container and start fresh when I replaced the ps yaml files (replacing and restarting the container did not work for me).

dereknop:
> Sorry I didn't test this before, but I got a crash on the progress table when I dropped down the `Type` pulldown. Also, the text scenario columns are all IO, MJ, QOL, etc.

Got it, working on it

brianpippin:
We need to tie a bow on this, so long as things don't break we need to get this deployed so we can test some before data collection tomorrow.  Everyone hop on it.

PR #301, opened 2025-06-19
Dev to Main

brianpippin:
* text result fix

* headers and comp fix

* take survey reactivtaed

* gen comparison. page loading baseline logic

* multi kdma, block randomization

* assigned on alignment

* only aligned for multi kdma

* determine multi kdma group and comparison accordingly

* refactor

* record aligned misaligned and baseline in surveyResults

* overlap

* fixed overlap logic

* clean

* fix group, review delegation sorted

* online only setup, return url

* consent form

* survey link on participant progress page

* missing commas

* complete button override

* copy editor


nextcen-dgemoets:
FYI, I had to delete my containers and rebuild them.  Before that, I got a message about not being able to find `./OnlineOnly/consentForm2025.pdf`.

And maybe this has been around for a while, but is this a problem?
`survey-core has version '1.12.30' and survey-react-ui has version '1.12.31'. SurveyJS libraries should have the same versions to work correctly.`

nextcen-dgemoets:
> I also can't reproduce that. Where is that being logged for you @nextcen-dgemoets? Docker logs or console?

Dev console.

joshuale-caci (approved):
Looks good on my end! Got the consent form to appear and did not have any error logs as well.

nextcen-dgemoets:
The only other "anomaly" I'll note is that this time, after I finished the delegation survey, it gave me the thank you page, then directed me to the prolific login page.  Last time, it took me to the thank you page then popped up the "Do you want to leave" dialog.   Which is the correct behavior?
 
The only difference I can think of is that I had logged myself out of the dashboard before starting the survey this last time.  Not saying that made a difference, though.

brianpippin:
> The only other "anomaly" I'll note is that this time, after I finished the delegation survey, it gave me the thank you page, then directed me to the prolific login page. Last time, it took me to the thank you page then popped up the "Do you want to leave" dialog. Which is the correct behavior?
> 
> The only difference I can think of is that I had logged myself out of the dashboard before starting the survey this last time. Not saying that made a difference, though.

I have been doing it not signed every time as that is what the user should be doing, and I get the thank you and then straight back to prolific, which I believe is the desired behavior.  The being logged in is probably why you get the do you want to leave message, which is fine because that is logged in behavior where it is redirecting.  IMO, I think this is ok.

nextcen-dgemoets (approved):

PR #302, opened 2025-06-24
Itm 1042

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&jql=assignee%20IN%20%2860ba425da547eb00686ee0ce%2C%20empty%29&selectedIssue=ITM-1042)

A lot of funkiness with the progress table should now be cleaned up:

- Phase 2 table should now have a del-5 column
- Text based scenario start and end times are properly being pulled (before it was grabbing the start and end times from the same scenario)
- The .xlsx download should now separate the two phases and only include the headers for that phase. These column headers should be in the same order as in the UI and the highlighting behavior should carry over into the download
- You shouldn't see any phase 1 fields (e.g `MJ-1`) in the phase 2 data and vice versa

### DIFF
--- a/dashboard-ui/src/components/Account/participantProgress.jsx
+++ b/dashboard-ui/src/components/Account/participantProgress.jsx
@@ -64,7 +64,14 @@ export function ParticipantProgressTable({ canViewProlific = false }) {
 
     const getCompletionOptions = () => {
         const textThreshold = selectedPhase === 'Phase 2' ? 4 : 5;
-        return [`All Text (${textThreshold})`, 'Missing Text', 'Delegation (1)', 'No Delegation', 'All Sim (4)', 'Any Sim', 'Adept + OW Sim', 'No Sim'];
+        const baseOptions = [`All Text (${textThreshold})`, 'Missing Text', 'Delegation (1)', 'No Delegation', 'All Sim (4)', 'Any Sim', 'No Sim'];
+        
+        if (selectedPhase === 'Phase 1') {
+            // phase 1 option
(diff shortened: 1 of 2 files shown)


jaudick (approved):
The ui and downloaded csv tables all look as described

nextcen-dgemoets (commented):
It looks like you also changed the Download Filtered Data to be local to a specific Phase, which is good!  It's only present if you've filtered the current Phase of data, whereas before data was always filtered because you either had Phase 1 or Phase 2 selected.  That said, I wonder if the "Showing x of y rows based on filters" line should change accordingly-- that is, if you're looking at unfiltered Phase 1 (or 2) data, this line doesn't show.  Then if you apply a filter, the denominator is the count for the current phase.

Also, not part of this PR I guess, but the "All Text (5)" filter should be changed to 4 for Phase 2, no, since that's the max?

dereknop:
> It looks like you also changed the Download Filtered Data to be local to a specific Phase, which is good! It's only present if you've filtered the current Phase of data, whereas before data was always filtered because you either had Phase 1 or Phase 2 selected. That said, I wonder if the "Showing x of y rows based on filters" line should change accordingly-- that is, if you're looking at unfiltered Phase 1 (or 2) data, this line doesn't show. Then if you apply a filter, the denominator is the count for the current phase.
> 
> Also, not part of this PR I guess, but the "All Text (5)" filter should be changed to 4 for Phase 2, no, since that's the max?

Agree on both fronts, updated

nextcen-dgemoets (approved):
LGTM!

PR #306, opened 2025-07-04
Preliminary ITM-1044 RQ1/3/4 Changes

brianpippin:
Initial Shot at ITM-1044, still needs work and some checking from Derek probably.


brianpippin:
From Jennifer if looking at this:

In the RQ1/3 dataset you sent, it is looking better but I'm not getting any of the trust rating trials for the AF-MF DMs. It looks like it is only showing the comparison trial, but with no data.
As you know, the multi-KDMA trust and delegation is the new finding that we are most interested in seeing from this dataset so I need those asap.
I pushed Audrey off until Wednesday morning but I promised we would have everything ready then.

I am also concerned about two of the columns: Alignment score (Delegator\|Observed_ADM (target))
is not matching with ADM_aligned_status as expected, aligned should have the highest alignment score most of the time and misaligned should have the lowest. They seem all over the place. 
Possibly related: ADM Loading has "exemption" for every single aligned and misaligned ADM....that can't be right.

I also need to report the alignment for the 3 sets we ran for the observation materials but it looks like those are all mixed in or not reported on the dashboard. The Alignment (ADM|Target) column is concerning because I didn't think the alignment for any of the aligned ADMs was as low as .43 or .41 like it is in that column

dereknop (approved):

nextcen-dgemoets (commented):
Is there another ticket for updating the eval 8 definitions spreadsheet/table?  Or are we doing it for this PR?  Because there are a number of edits we'll need there at some point. If we want them fixed here, I'll list them out.

Also, we could remove the TA1 and TA2 filters.  

The `Delegator_grp` filter only has Civilian and Military-- should it have "emailParticipant" and "Online"?  It doesn't in Phase 1 either, hmm.

I also got a crash when dropping down the `Scenario` filter on RQ1 and RQ3:
dash-error

dereknop:
For now, I am going to leave `Delegator_grp` and `Delegator_mil` in because idk if we will want to do something with that in the future. Removed TA1 and TA2 filters for phase 2 and fixed a bug for crashing on PID search. Working on the fix for the crash you described @nextcen-dgemoets

nextcen-dgemoets:
> For now, I am going to leave `Delegator_grp` and `Delegator_mil` in because idk if we will want to do something with that in the future.

Should `Delegator_grp` have "emailParticipant" and "Online" so that the filter at least works for something?  I'm not sure why it's not in the Phase 1 version of the toggle, so maybe that's deliberate?

nextcen-dgemoets:
Some things:
- The `Delegator_grp` toggle only has Civilian and Military, but the definition lists "Military, civilian, emailParticipant, Online".
   - And 'civilian' should be capitalized
- PID `202506159` has a `Delegator_Role` of "doctor" (lowercase), but this isn't one of the choices listed in the definitions.  Maybe they selected "Other" in the form and wrote in "doctor"?
   - If so, maybe the def should include "Other" as a catch-all for any fill-ins.
- In the definitions, the "SS" attribute is "Search or Stay," not just "Search".
- "Comparison" ADM types are in the table, but not listed in the definitions.
- Definition of `ADM Loading` says "Data is from the Phase 1 server alignments."
- Out of curiosity, why are there no rows with affiliation 0.4,6,7,8,9 or merit 0.0,1,4,6.9?  All other targets are represented in the data.

dereknop:
Fixed. 

Regarding your last question, there are several factors to consider. For example, 0.0 and 0.1 for Merit overlap with the baseline for MF2 and MF1, so they will never be shown. Our sample size may not be big enough yet that a participant has been most aligned or least aligned to those targets. And/or that could also be a product of the weighting of the text-based scenarios, with only 6 probes assessed per attribute for the participant, causing certain targets to be more commonly aligned to than others. The overlap in targets is laid out in `getScenarioGroups` in `/Survey/surveyUtils.js`. As far as I have tested, and have heard from everyone else that tested, the most aligned and least aligned targets are properly loading in.

dereknop (approved):

nextcen-dgemoets (approved):
Totally my mistake, but it's "Search vs. Stay", not "Search or Stay".

PR #307, opened 2025-07-07
ITM 1014: Resolving ESLint Build Warnings

NeilDaniel07:
**Description:** Previously, when the dashboard-ui was built, 863 warnings were generated by the linter. These included eqeqeq strict vs. loose equality warnings, react hooks missing components of the dependency array, unused variables, unclear order of operations in logical expressions, and more. This resolves all 863 warnings generated by ESLint. The most significant changes include the following: 

- `==` comparisons have been modified to `===` to ensure equality of type is enforced in comparisons
- In required cases where types do not match, comparison statements now use type coercion
- Missing react hook dependencies have been added to relevant dependency arrays
- Functions have been wrapped in React callbacks for memoization of state whenever possible
- Unused variables have been eliminated or commented out
- Redundant constructors that replicated default constructor behavior have been eliminated
- Unsafe variable access in inline functions within loops was eliminated

**Methodology:** Testing was primarily conducted manually through UI regression testing on components that were modified. This was because many runtime errors were not caught by the automated test suite. 

**(Additional) Bug Fixes:** 
1.  Previously, in production, when the user visited this page (https://darpaitm.caci.com/participant-progress-table), modified the `Sort By` parameter to be empty, and then pressed the Refresh Data button, the following error would be encountered: image This was due to failure to account for an edge case in a data sorting function. This bug is now resolved, with no sorting being applied to the rows of the table if the `Sort By` parameter is empty.
2. In production, when the user visited the RQ3 data table (http://localhost:3000/research-results/rq3) and attempted to filter data using the `Probe Set Assessment` or `Probe Set Observation` columns, a runtime warning occured:image. This is patched in this PR, without any changes to the functionality of data filtering in the table or downloaded Excel file.

**How To Test:** To verify all ESLint warnings have been resolved, cd/navigate into the dashboard-ui subfolder of the repository. Using PowerShell, run the following terminal command: `npm run build 2>&1 | Out-File -FilePath lint-warnings.txt -Encoding UTF8`. Alternatively, the equivalent Command Prompt command is `npm run build > lint-warnings.txt 2>&1.` A file titled `lint-warnings.txt` will be generated in the folder. Open this file and verify that no warnings for any particular lines in the repo were generated. For reference, the corresponding file of the original 863 warnings is linked here: 
[original-warnings.txt](https://github.com/user-attachments/files/21207560/original-warnings.txt). Then, visit the dashboard at http://localhost:3000/ and navigate through the pages. Ensure that the original functionality is fully maintained (there are no functional differences from the dashboard produced by the development branch). Finally, test the additional bug fixes. First visit http://localhost:3000/participant-progress-table, setting the `Sort By` parameter to empty, and pressing the `Refresh Data` button. Ensure that a runtime error is not generated and the data is correctly displayed in the table. Next, navigate to http://localhost:3000/research-results/rq3 and open your console. Attempt to filter the data on both the `Probe Set Assessment` and `Probe Set Observation` columns. Ensure that no runtime warning is generated and that data filtration works as expected in both the table and downloaded Excel file.

**Follow Up:** There is still a warning generated since the `crypto` module cannot be resolved. This seems to be required for hashing. However, it's unusual that the build still compiles without the module being imported. It likely indicates that the logic using this module is never triggered by typical user action. I recommend we disable the fallback in a polyfill for this module. I believe this would require switching from `react scripts` to `react-app-rewired` to build. We should determine what changes, if any, need to be made. 

**Follow Up:** There is still a warning generated since the `crypto` module cannot be resolved. This seems to be required for hashing. However, it's unusual that the build still compiles without the module being imported. It likely indicates that the logic using this module is never triggered by typical user action. I recommend we disable the fallback in a polyfill for this module. I believe this would require switching from `react scripts` to `react-app-rewired` to build. We should determine what changes, if any, need to be made.


brianpippin:
> I appreciate the feedback. The reason I chose to opt for explicit casting in a lot of places throughout the codebase is because, in many cases, current logic was relying on type coercion/conversion. Many comparisons were thus failing after switching to `===`, because types didn't match (number vs. string and string vs. undefined were two common occurrences). I've provided a couple specific examples below of where simply changing `==` to `===` and `!=` to `!==` broke existing logic and led to missing UI components or runtime errors. Additionally, reading through JSX and JS files while trying to keep track of the expected type of function parameters across complex logic was somewhat confusing. I figured that type coercion adds some implicit documentation regarding the expected type of parameters used throughout the function. If the concern is that type casting adds unnecessary overhead, my understanding is that casting a number or nullish type to a String is a constant time operation. Most Strings that might be redundantly cast to Strings are quite small in length (ex: pid or fields of a JSON object), so I don't believe linear time conversion here is substantial. Let me know if there are specific areas where it would make more sense to inspect the type of parameters individually and ensure `===` comparisons are safe to avoid type casting. Thanks!
> 
> **Example One (Line 410 of survey.jsx file in src/components/survey):** ` else if (this.state.surveyVersion == 1.3 || this.state.surveyVersion == 1.0) {` Changing `==` to `===` results to the text survey producing runtime warnings. This is because the surveyVersion parameter is a String.
> 
> **Example Two (Line 261 of utils.js in src/components/Research):** `const scenarioMatches = obj['scenarioIndex'] == scenario;` Scenario is a boolean type but may change types to a String depending on some earlier logic. Enforcing a strict `===` type is unpredictable because the type of the scenario variable is not known until runtime.

So I did want strong typing but I think some of what we are doing is backwards, so take this example:

from admChart.js:

} else if (this.state.currentEval.value == 3) {
} else if (Number(this.state.currentEval.value) === 3) {

This is not what doing strong typing should be, I would prefer this was:

} else if (this.state.currentEval.value === '3') {

We shouldn't be doing Number or String on a variable when we are comparing it to a static property, we should make the static property the appropriate type.   I am also curious that we are doing number and string on almost every "===" now.  DId we check to see if that was needed.  I am now worried about how so many underlying variables are probably mistyped, or did we just add String and Number along with every "===" as a guess that they were needed to work.  We need to be careful because, I know it doesn't seem like much but all these extra function calls can add up on performance in some places.  I think I had envisioned trying to move everything to "===" but we would need to examine every case to make sure, because usually when the triple equals doesn't work something else is amiss.

jaudick:
@NeilDaniel07 
I wasn't thinking about overhead and time complexity but its great you are thinking about those. They may add up but I have no idea how much of a performance hit it is.
 
I was more worried about maintaining the code as devs and what these types actually are. As Brian said:

> We shouldn't be doing Number or String on a variable when we are comparing it to a static property, we should make the static property the appropriate type
  
and

> Usually when the triple equals doesn't work something else is amiss.

---

@brianpippin
You said:
> I am now worried about how so many underlying variables are probably mistyped

I don't totally understand what Neil should do then. This sounds like he needs to track down mistyped variables. Double equals `==` seemed to be handling the int and string issues and pushing that under the rug. What should he do about the casting?

brianpippin:
> @NeilDaniel07 I wasn't thinking about overhead and time complexity but its great you are thinking about those. They may add up but I have no idea how much of a performance hit it is.
> 
> I was more worried about maintaining the code as devs and what these types actually are. As Brian said:
> 
> > We shouldn't be doing Number or String on a variable when we are comparing it to a static property, we should make the static property the appropriate type
> 
> and
> 
> > Usually when the triple equals doesn't work something else is amiss.
> 
> @brianpippin You said:
> 
> > I am now worried about how so many underlying variables are probably mistyped
> 
> I don't totally understand what Neil should do then. This sounds like he needs to track down mistyped variables. Double equals `==` seemed to be handling the int and string issues and pushing that under the rug. What should he do about the casting?

I am thinking we don't do any typing unless we know that that switching to the "===" causes a crash, and then we fix just those crashes by checking the types.  I can't imagine there is a ton of them (or I hope not), but by only changing the "==" to "===" then building, we should be able to troubleshoot the issues that arise 1 by 1, which could lead us to any underlying typing issues.  Looking through the code review, it just looks like every "==" to "===" was just typed automatically to match, but I could be wrong.

NeilDaniel07:
> 
> I am thinking we don't do any typing unless we know that that switching to the "===" causes a crash, and then we fix just those crashes by checking the types. I can't imagine there is a ton of them (or I hope not), but by only changing the "==" to "===" then building, we should be able to troubleshoot the issues that arise 1 by 1, which could lead us to any underlying typing issues. Looking through the code review, it just looks like every "==" to "===" was just typed automatically to match, but I could be wrong.

I just pushed a commit that takes this approach. `===` is used whenever the types of variables being compared are the same and type coercion is used only in cases where casting is required. Since the development branch has been modified recently, the quantity and locations of linter warnings has been changed. Please re-download the` original-warnings.txt `file (the updated version is available in my original PR request above). I've also identified and fixed an additional bug in this commit - details on this and how to test that its been resolved is also available in the original PR comment.

dashboard-ui/src/components/Account/participantProgress.jsx
```diff
@@ -104,21 +132,21 @@ export function ParticipantProgressTable({ canViewProlific = false }) {
...
                 const survey_end_date = new Date(lastSurvey?.results?.timeComplete);
-                obj['Del Start Date-Time'] = survey_start_date != 'Invalid Date' ? `${survey_start_date?.getMonth() + 1}/${survey_start_date?.getDate()}/${survey_start_date?.getFullYear()} - ${survey_start_date?.toLocaleTimeString('en-US', { hour12: false })}` : undefined;
-                obj['Del End Date-Time'] = survey_end_date != 'Invalid Date' ? `${survey_end_date?.getMonth() + 1}/${survey_end_date?.getDate()}/${survey_end_date?.getFullYear()} - ${survey_end_date?.toLocaleTimeString('en-US', { hour12: false })}` : undefined;
+                obj['Del Start Date-Time'] = String(survey_start_date) !== 'Invalid Date' ? `${survey_start_date?.getMonth() + 1}/${survey_start_date?.getDate()}/${survey_start_date?.getFullYear()} - ${survey_start_date?.toLocaleTimeString('en-US', { hour12: false })}` : undefined;
+                obj['Del End Date-Time'] = String(survey_end_date) !== 'Invalid Date' ? `${survey_end_date?.getMonth() + 1}/${survey_end_date?.getDate()}/${survey_end_date?.getFullYear()} - ${survey_end_date?.toLocaleTimeString('en-US', { hour12: false })}` : undefined;
```
jaudick, line 149: Want to confirm this is what's happening here.
`survey_start_date` is of type `Date()` and not a `string`. And you need to cast the `Date()` type to a `string`?

NeilDaniel07: Yes, that's correct. Only using `===` will cause the comparison to fail. The `.toString()` method that will be invoked by casting `survey_start_date` to a `String` will represent null/undefined/badly formatted dates as 'Invalid Date' in a `String` format, which is what causes the comparison to work as expected.

NeilDaniel07:
> Why are so many non code files showing that they have been changed in the files changed area?

I believe this is because I rebuilt the dashboard-ui using `npm run build`, which automatically generated these new non-node files in the build folder. I noticed that the development branch currently doesn't have the build folder so I assume it's been manually removed  (I didn't see the folder in `.gitignore` - I can add it if you would like). I can delete the build folder for this PR in a new commit if we don't want to track build artifacts - just let me know.

brianpippin:
> > Why are so many non code files showing that they have been changed in the files changed area?
> 
> I believe this is because I rebuilt the dashboard-ui using `npm run build`, which automatically generated these new non-node files in the build folder. I noticed that the development branch currently doesn't have the build folder so I assume it's been manually removed (I didn't see the folder in `.gitignore` - I can add it if you would like). I can delete the build folder for this PR in a new commit if we don't want to track build artifacts - just let me know.

We definitely do not want the build folder checked in, so adding to gitignore and removing from this PR would be useful.  We don't run npm run build, because docker will build the container for us.

NeilDaniel07:
> > > Why are so many non code files showing that they have been changed in the files changed area?
> > 
> > 
> > I believe this is because I rebuilt the dashboard-ui using `npm run build`, which automatically generated these new non-node files in the build folder. I noticed that the development branch currently doesn't have the build folder so I assume it's been manually removed (I didn't see the folder in `.gitignore` - I can add it if you would like). I can delete the build folder for this PR in a new commit if we don't want to track build artifacts - just let me know.
> 
> We definitely do not want the build folder checked in, so adding to gitignore and removing from this PR would be useful. We don't run npm run build, because docker will build the container for us.

That makes sense. The reason I chose to use `npm run build` for this ticket is that it made it much easier to pipe the warnings printed in the terminal to an out file. I'll update `.gitignore` in my next commit to resolve merge conflicts.

NeilDaniel07:
> > Why are so many non code files showing that they have been changed in the files changed area?
> 
> I believe this is because I rebuilt the dashboard-ui using `npm run build`, which automatically generated these new non-node files in the build folder. I noticed that the development branch currently doesn't have the build folder so I assume it's been manually removed (I didn't see the folder in `.gitignore` - I can add it if you would like). I can delete the build folder for this PR in a new commit if we don't want to track build artifacts - just let me know.

I added the build folder to gitignore - please let me know if there are any other non-code files that have been modified that you weren't expecting.

nextcen-dgemoets (commented):
Looks good to me.  Somebody who knows the dashboard / React better should look at the code more closely, but I didn't encounter any functional issues or lint warnings other than the `crypto` one noted in the PR.  The Progress Table error was reproduced on prod and was fixed in this branch.  I noted one runtime warning, but it happens on prod as well so at a minimum it doesn't need to be fixed as part of this ticket, but it DOES have to do with int / string issues, so maybe it's tangentially relevant.

But other than that, I'm inclined to "approve".

brianpippin:
> > Looks good to me. Somebody who knows the dashboard / React better should look at the code more closely, but I didn't encounter any functional issues or lint warnings other than the `crypto` one noted in the PR. The Progress Table error was reproduced on prod and was fixed in this branch. I noted one runtime warning, but it happens on prod as well so at a minimum it doesn't need to be fixed as part of this ticket, but it DOES have to do with int / string issues, so maybe it's tangentially relevant.
> > But other than that, I'm inclined to "approve".
> 
> @brianpippin and @dereknop, would you like me to investigate and fix the runtime warning Darren found as part of this PR?

I would at least investigate.  If it looks time consuming make another ticket, if not a hard fix, add to this.

dashboard-ui/src/components/Survey/survey.jsx
```diff
@@ -132,23 +132,29 @@ class SurveyPage extends Component {
         this.uploadButtonRef = React.createRef();
         this.uploadButtonRefPLog = React.createRef();
         this.redirectLinkRef = React.createRef();
-        if ((this.state.surveyVersion == 4.0 || this.state.surveyVersion == 5.0 || this.state.surveyVersion == 6.0) && this.state.pid != null) {
+
+        console.log(typeof(this.state.surveyVersion));
+        console.log(this.state.surveyVersion);
```
brianpippin, line 137: These console log statements should be removed.

dashboard-ui/src/components/ReviewTextBased/ReviewTextBased.jsx
```diff
@@ -35,8 +35,9 @@ export function ReviewTextBasedPage() {
     const textBasedConfigs = useSelector(state => state.configs.textBasedConfigs);
     const [selectedConfig, setSelectedConfig] = useState(null);
 
+    //Function is left in case future modifications are required to enable behavior on survey completion
     const handleSurveyComplete = (sender) => {
-        const results = sender.data;
+        //const results = sender.data;
```
brianpippin, line 40: I would prefer we remove the comment about leaving the function in, remove the commented out code, and put a comment inside the function to say /* Add survey completion functionality here */ or something similar.

dashboard-ui/src/components/ReviewDelegation/ReviewDelegation.jsx
```diff
@@ -82,8 +82,9 @@ export function ReviewDelegationPage() {
     const [selectedConfig, setSelectedConfig] = useState(null);
     const [reviewingText, setReviewingText] = useState(null);
 
+    /* Function is left in case future changes are reuqired to the survey completion behavior */
     const handleSurveyComplete = (sender) => {
```
brianpippin, line 86: I would prefer we remove the comment about leaving the function in, remove the commented out code, and put a comment inside the function to say /* Add survey completion functionality here */ or something similar.

dereknop: I think we can just delete this function and the reference to it on line 267. We are using the default behavior here, and we don't need to take any action with someone's answers from the review delegation page, so I don't see a use for this function.

dashboard-ui/src/components/ReviewTextBased/ReviewTextBased.jsx
dereknop: Ditto what I said below

dashboard-ui/src/components/OnlineOnly/OnlineOnly.jsx
```diff
@@ -18,6 +18,7 @@ const ADD_PARTICIPANT = gql`
         addNewParticipantToLog(participantData: $participantData, lowPid: $lowPid, highPid: $highPid) 
     }`;
 
+/*
```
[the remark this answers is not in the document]
brianpippin: > Depends on if we will be doing an online only data collection using the phase 1 materials at some point in the future

We will be doing a phase 1 data collect in the UK, but if this is needed for Phase 1 but not anymore there should be a config option for this or attached to another config option so we don't need to rebuild to roll back to Phase 1.

brianpippin:
Great job, I left some minor comments, but when they are all resolved/taken care of, this looks good to me.

dashboard-ui/src/components/Survey/survey.jsx
NeilDaniel07: Sorry for missing those, they should be removed now.

dashboard-ui/src/components/Survey/survey.jsx
```diff
@@ -132,23 +132,29 @@ class SurveyPage extends Component {
...
+
+        const surveyVersionNumber = Number(this.state.surveyVersion);
```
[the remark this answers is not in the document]
NeilDaniel07: It is a String type. I updated the comparison statements so that `this.state.surveryVersion` would be compared to String types instead of numbers and tested the survey page to ensure that no functionality broke.

dashboard-ui/src/components/Survey/dynamic.jsx
```diff
@@ -207,6 +207,7 @@ const Dynamic = ({ patients, situation, supplies, decision, dmName, actions, sce
         return processedText;
     };
 
+    /*
```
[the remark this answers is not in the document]
NeilDaniel07: I removed it in the latest commit.

dashboard-ui/src/components/OnlineOnly/OnlineOnly.jsx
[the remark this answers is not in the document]
NeilDaniel07: I added a config file at `src/components/OnlineOnly/config.js` that contains the `LOG_VARIATIONS` object as well as a boolean variable representing whether it's in use (which I set to false for now). I also removed the reference to the object in the `OnlineOnly.jsx` file (to prevent an unused variable linter warning). The config file can be imported if needed and the object can be loaded conditionally depending on the value of this boolean. Please let me know if this works.

NeilDaniel07:
> Great job, I left some minor comments, but when they are all resolved/taken care of, this looks good to me.

I appreciate your thorough feedback. I think I've taken care of all the concerns you raised in the latest two commits, but please let me know if something looks incorrect. In particular, if you could take a look at the way I set up the config file to address your last issue regarding the unused array object and make sure that it's what you were expecting, that would be great. Thank you.

NeilDaniel07:
> We should also make sure we run the test suite before merging this to make sure no tests break.

For some reason, multiple tests are failing for me (but these same tests also fail on the development branch). Could someone else please try running the test suite to make sure nothing has broken in this branch? I'm not sure if my tests are failing because of some package issues (I spent a lot of time resolving dependencies and this may have broken something).

nextcen-dgemoets:
> > We should also make sure we run the test suite before merging this to make sure no tests break.
> 
> For some reason, multiple tests are failing for me (but these same tests also fail on the development branch). Could someone else please try running the test suite to make sure nothing has broken in this branch? If so, that would be amazing. I'm not sure if my tests are failing because of some package issues (I spent a lot of time resolving dependencies and this may have broken something).

I checked both `development` and this branch, and (especially after fixing/updating a few tests), the exact same tests failed for the same reason.  So I don't think this work introduced any new tests issues/failures.

PR #311, opened 2025-07-10
ITM 1050: Probe IDs Column Added To RQ 2.2/2.3 Table

NeilDaniel07:
**Original Ticket:** [Click Here](https://nextcentury.atlassian.net/browse/ITM-1050?atlOrigin=eyJpIjoiMWNlYmNhODhlYWM2NGIwOGIwNDUxMzIyOGY1ZDEzNGQiLCJwIjoiaiJ9)

**Description:** This implements changes to the RQ 2.2/2.3 table in the dashboard. The table now contains an additional column named `Probe IDs`. This shows the randomly generated Probe IDs corresponding to the particular set of a row, sorted by the attribute first (in the case where probes from multiple attributes are present in a single set - for a multiKDMA scenario). If the attribute for all probes are the same, then sorting is performed based on the probe number. The changes made to the table are also visible in the Excel download, where a new column called `Probe IDs` contains the probe ID information. The Probe IDs are visible for both synthetic and observation sets. Additionally, the Variable Definitions table and corresponding Excel file have been updated to provide Source, Definition, and Levels information for this new `Probe IDs` variable.

**Server Schema Changes:** To support the query for sourcing probe information, modifying the server schema was required. The `getAllHistoryByEvalNumber` projection now includes the `synthetic`, `scenario`, `alignment_target`, `probe_ids`, and `history.parameters.probe_id` fields. This does not have any downstream impact on other components which use this projection. 
 
**Disclaimers:**
- Originally, the set of Probe IDs was supposed to appear in a popup when the user clicked a button in the Set column. The approach of adding an additional column to the table was an alternative to this that was determined to be more optimal. The reason for this is that the Downloaded Excel file will now have the Probe ID information.
- Probe information is not available for every document in the `admTargetRuns` collection of the database. It is available for every document that contains the `synthetic` parameter with a corresponding parameter of `true`. While all rows in the table have available probe information, it's important to note that in the case that no probe information is available, the appropriate cell of the table/generated Excel file will remain empty. 

**(Additional): Sorting Improvements:** Previously, once sorted primarily on attribute, rows of the table were sorted by using default JS string comparison, which meant `P2June Dynamic Set 100` appeared before `P2June Dynamic Set 2`. This is resolved by using a custom Regex based numeric extractor to sort rows by set numbers in ascending order.

**How To Test:**
1. Navigate to the RQ 2.2/2.3 table on the RQ2 page of the Data Analysis tab. http://localhost:3000/research-results/rq2. Make sure that Phase II June 2025 Collaboration is selected at the top of the page.
2. Ensure that a new Probe IDs column appears to the right of the Set column. 
3. Check that this column contains Probe IDs in the agreed upon `Probe-AttributeType-Number` format for each row, including multiKDMA scenarios and rows corresponding to observation sets. Ensure Probe IDs are sorted in the previously described manner. 
4. Make sure that filtering based on attribute/set/etc. still works as before.
5. Check to see that for each attribute, the table is now rendered with rows sorted by ascending Set number by viewing the Set column. 
6. Download the Excel file to verify that it includes the “Probe IDs” column with the same formatted values.
7. Click on the `View Variable Definitions` button at the top right of the table container. Check that a new column titled `Probe IDs` is visible next to the Set column with populated Source/Definition/Level information. Download this file to make sure that the same structure is present in an Excel format.


NeilDaniel07:
> You need to merge in the latest from development into your branch, because that table has been updated since you started this PR. (That's why there are conflicts that must be resolved). For example, the set names are now e.g. "P2June Dynamic Set 23" and "P2June Observation Set 2".

Apologies for not getting to this last night with my initial PR. All merge conflicts should now be resolved in the latest commit.

nextcen-dgemoets:
I was able to verify by comparing to the DB that the correct probes are in the new column.  However, it looks like you're sorting the probes numerically, which normally makes sense but we may not want to do that here:  the probes were in "bin" order, meaning the first listed probe was in bin 1 (lowest kdma value) and the last probe was in bin 6 (highest kdma value).  We should check with Jennifer, but they might want to be able to know the bin from the probe ID...but not sure.  If not, then keep it sorted numerically.

It looks like the P2 June Observation Sets (1-3) are now missing from the `Set` dropdown (compare to [prod](https://darpaitm.caci.com/research-results/rq2) to see what I mean).  Maybe this was something that happened in the merge conflict resolution?

It also occurs to me that until/unless we get TA1 to agree with our probe id naming convention-- and update their probe ids-- we are technically not listing the probe IDs in this column (as the column title suggests).  Maybe the column name should just be "Probes".  That said, maybe it doesn't matter.  Does @brianpippin  have an opinion on this?

Once we've finalized the column title, the variable definitions file (Excel doc) will need to be updated for this new field.  For example:
- `Source`:  `TA1`
- `Definition`:  `The scenario probe IDs selected by TA1 for the Observation set, or randomly selected from TA1-defined "bins".  IDs are in bin order.`
  - or `in numerical order`, if that is the decision.
- `Levels`:  `Probe-<kdma>-<number>`

Run this definition by Jennifer for editing.  The above is just a draft I came up with.

nextcen-dgemoets:
Jennifer says bin order isn't important in RQ2, so maybe:
- `Source`:  `TA1`
- `Definition`:  `The scenario probe IDs selected by TA1 for the Observation set, or randomly selected from TA1-defined "bins".  IDs are in numerical order per KDMA.`
- `Levels`:  `Probe-<kdma>-<number>`

nextcen-dgemoets:
Everything looks good, except that there are three extra columns added to the end of both the table and the definitions that should be removed:
- `Scenario`, `alignmentTarget`, and `Synthetic`

nextcen-dgemoets (commented):
LGTM, everything has been addressed.  I could validate the values against the DB.

Just one last question below.

nextcen-dgemoets (approved):
All good!

dereknop (approved):
LGTM. Table populating as you described. Variable defs and download look as expected. Good work.

joshuale-caci (approved):
Probe Ids match up and everything looks correct

PR #313, opened 2025-07-16
Itm 1047

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?jql=assignee%20%3D%2060ba425da547eb00686ee0ce&selectedIssue=ITM-1047)

Note the new probe naming convention. Go to the `Participants` tab for any phase 2 scenario. You should now see three download options. 
Screenshot 2025-07-16 at 11 50 50 AM
Scenario-specific data, attribute-wide (all 3 scenarios in that group), and all data across all scenarios. 

Please spot-check that the data in the download accurately reflects the participant's data.

### DIFF
--- a/dashboard-ui/src/components/TextBasedResults/TextBasedResultsPage.jsx
+++ b/dashboard-ui/src/components/TextBasedResults/TextBasedResultsPage.jsx
@@ -142,7 +142,7 @@ function getQuestionTextLegacy(qkey, scenario, textBasedConfigs) {
 }
 
 
-function ParticipantView({ data, scenarioName, textBasedConfigs, selectedEval }) {
+function ParticipantView({ data, scenarioName, textBasedConfigs, selectedEval, participantBased, scenarioOptions, participantLog }) {
     const [organizedData, setOrganizedData] = React.useState(null);
     const [excelData, setExcelData] = React.useState(null);
     const [orderedHeaders, setHeaders] = React.useState([]);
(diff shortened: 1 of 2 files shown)


NeilDaniel07 (approved):
Everything looks good to me. The new Probe ID format matches the requirements of the ticket and what was Jennifer expressed she wanted in the Dashboard channel. All three download button options appear for me for each scenario in the `June 2025 Collaboration` tab. Scenario-specific data download matched the expected column names/order and format of the table in the dashboard. 

My only comment is on the order of columns in the attribute wide data download. While I was able to verify that all 36 columns for the three scenarios in each attribute group were present, I saw that the order of columns was seemingly arbitrary. It did not seem to be grouped in the same order as columns within a scenario nor by scenario in general. However, the order does seem to be deterministic (consistent across multiple file downloads). I doubt this is an issue so I'm approving for now, but I just wanted to put it on your radar out of extra caution.

nextcen-dgemoets (approved):
I didn't see any problems either, and did some spot-checking of results/consistency.  I agree that the probe order in the download is confusing.  At first I thought it was in "bin order" but I don't think it is.  Do you now how column order is determined?

But like @NeilDaniel07 , I can still approve it in its current form.

dereknop:
Yea good point @NeilDaniel07  @nextcen-dgemoets,  I'll fix that.

nextcen-dgemoets:
> Yea good point @NeilDaniel07 @nextcen-dgemoets, I'll fix that before merging.

I'm not seeing a difference in the Download Scenario Results download (e.g., `text_result_data_June2025-AF3-eval.xls`).

The others I'm seeing a difference but I'm still trying to reconcile them.
The "All AF" xls sorts its probes as 4, 17, 14, 10, 1, 18, 35, 37, 24, 33, 22, 44, 43, 45, 53, 52, 42, 54.

But the June AF bins are, I believe:
- Bin 1:  [18, 24, 52], Bin 2:  [17, 22, 53], Bin 3: [14, 35, 42], Bin 4: [10, 33, 54], Bin 5: [4, 44, 45], Bin 6: [1, 37, 43]

If instead of by bins, you were sorting them by assessment sets (which contain one probe from each bin), then you'd have
- Set 1: [18, 17, 14, 10,  4,  1], Set 2 [24, 22, 35, 33, 44, 37], Set 3 [52, 53, 42, 54, 45, 43]

So I guess I'm still not clear on the order.

dereknop:
> So I guess I'm still not clear on the order.

I am sorting by assessment sets. I've pushed a to make sure they are presented in the order they were shown in the text-based scenarios (I was sorting by set, but assuming the result document preserved that order, which it did not). 

There is an exception for the sets that had a probe change between the first set of June participants and the online-only data collection; in these cases, I tried to fit the old probe as close to where it would have been as possible. Using your example, my sets for AF are 

1: [1, 4, 10, 14, 17, 18]
2: [22, 24, 33, 35, 37, 44]
3: [42, 43, 45, 52, 53, 54]

That is the order they were presented in the text-based scenarios, and should now be what you see for a `Download All AF Results`. These orders should persist in the `Download All Results` as well.

PR #314, opened 2025-07-18
ITM 1054: ADM Loading Information Popup on Participant Progress Table

NeilDaniel07:
**Original Ticket:** [Click Here](https://nextcentury.atlassian.net/browse/ITM-1054?atlOrigin=eyJpIjoiZWQ2YmFiMTFmMDAzNDUwNmJjY2I4ZjFjNjI2ZmYxYzQiLCJwIjoiaiJ9)

**Description:** This PR regards changes to the Participant Progress table in the dashboard. The `Del-1` to `Del-5` columns on the `Phase 2` tab now render as buttons which trigger a popup titled `ADM Information`. These buttons are generated for each delegation scenario ID that a participant has completed. The left side of the popup shows the `Participant ID`, `Scenario ID`, `Attribute/Type`, and either the list of all alignments (from most to least aligned) for single kdma scenarios or `KDMA Scores` for multi kdma scenarios. In parentheses next to each of these alignments is the alignment score. The right side of the popup shows a table with four columns of ADM information. These columns are `Type`, `ADM Name`, `Target`, and `ADM Loading`. The `ADM Name` field refers to the name of the ADM used for that particular alignment `Type` in survey data collection for a particular participant/scenario ID combination. The `Target` column refers to the ADM target parameter corresponding to this participant ID and scenario ID combination. `ADM Loading` represents whether the value in the `Target` column matches the most aligned (for the Aligned row) or least aligned (for the Misaligned row) item in the `All Alignments` list on the left side. This is computed by a call to the function `determineChoiceProcessJune2025` located in `components/Research/utils.js` and is not implemented with local comparison logic, per Jennifer's request. 

**Disclaimers:**
- The cells in the `Target` and ADM `Loading` columns corresponding to the baseline `Type` are hardcoded to display values of `N/A`. This is because Jennifer requested omitting this information since responses to the baseline ADM don't vary from target to target and the target column is irrelevant to baseline measurements.
- Popup logic including what to display and population of cells in the table for multi kdma scenarios was performed by @dereknop. The descriptions I have provided in this PR correspond closer to single kdma scenarios. Please direct any clarifications on multi-kdma scenario specific items to him. 

**Example Popup Screenshot (Single KDMA):**
Screenshot 2025-07-18 115715

**Example Popup Screenshot (Multi KDMA):**
image

**Additional Changes To Delegation Column:**
Per Jennifer's request, the Delegation Column in this table is no longer a binary representation of whether the participant started the survey process (this change applies to both `Phase 1` and `Phase 2`). The column now reflects the number of unique scenario ID surveys completed (even if the overall scenario is not completed). Thus, for `Phase 1`, this is on a scale of 0-4. For `Phase 2`, this is on a scale of 0-5. The background color of the cell only appears green in color if the survey has been fully completed.

**How To Test:**
1. Select the Phase 2 tab on the Participant Progress tab: http://localhost:3000/participant-progress-table.
2. Ensure that buttons appear for every cell where `Delegation != 0`. Further, check that the number of buttons in each row is the same as the value in the `Delegation` column. 
3. Click on several buttons across multiple participant IDs/rows for both single kdma and multi kdma scenarios. Ensure that all fields are populated similar to the example images provided above. Make sure that you are able to scroll down the popup and the `Close` button in the top corner closes the display. 
4. Verify that the `Participant ID`, `Scenario ID`, and `Attribute` information is correct on the popup. This is all information that is available elsewhere in the table.
5. For single kdma scenarios, ensure that the list of all alignments is in fact sorted from highest to lowest alignment by inspecting the values in parentheses next to each individual alignment. 
6. To be extra thorough if you would like, choose a particular `Participant ID.` Open your Mongo Compass and navigate to the `userScenario` results collection. Use the filter `{ participantID: "<Your PID here>" }` to bring up the unique document for your participant. Ensure that the `mostLeastAligned` array in this document matches what is displayed on the popup. Next, navigate to the `surveyResults` collection and use the filter `{"results.pid": "<Your PID here>" }` to find the corresponding document. Click through the single medic documents to find the ones corresponding to the attribute in your popup. Make sure the `ADM Name` and `ADM Target` information match what you see in your popup for both the aligned and misaligned type. Then, find the comparison document for your attribute and make sure the `baselineName` parameter matches what you see in your popup for the baseline `ADM Name` cell.
7. Make sure that filtering based on `type/eval/progress/etc.` and sorting still work as before. 
8. Download the Excel file to verify that development of the popup has caused no information changes to the downloaded file.

### DIFF
--- a/dashboard-ui/src/components/Account/participantProgress.jsx
+++ b/dashboard-ui/src/components/Account/participantProgress.jsx
@@ -73,7 +73,8 @@ export function ParticipantProgressTable({ canViewProlific = false }) {
 
     const getCompletionOptions = () => {
         const textThreshold = selectedPhase === 'Phase 2' ? 4 : 5;
-        const baseOptions = [`All Text (${textThreshold})`, 'Missing Text', 'Delegation (1)', 'No Delegation', 'All Sim (4)', 'Any Sim', 'No Sim'];
+        const delThreshold = selectedPhase === 'Phase 2' ? 5 : 4;
+        const baseOptions = [`All Text (${textThreshold})`, 'Missing Text', `Delegation (${delThreshold})`, 'No Delegation', 'All Sim (4)', 'Any Sim', 'No Sim'];
 
         if (selectedPhase === 'Phase 1') {
             // phase 1 option
@@ -323,6 +324,8 @@ export function ParticipantProgressTable({ canViewProlific = false }) {
     React.useEffect(() => {
         if (formattedData.length > 0) {
             const textThreshold = selectedPhase === 'Phase 2' ? 4 : 5;
+            const delThreshold = selectedPhase === 'Phase 2' ? 5 : 4;
+
             setFilteredData(formattedData.filter((x) => {
                 const participantPhase = getParticipantPhase(x);
                 const shouldShowInPhase = selectedPhase === `Phase ${participantPhase}`;
@@ -336,7 +339,7 @@ export function ParticipantProgressTable({ canViewProlific = false }) {
                     (evalFilters.length === 0 || evalFilters.includes(x['Evaluation'])) &&
                     (!completionFilters.includes(`All Text (${textThreshold})`) || x['Text'] >= textThreshold) &&
                     (!completionFilters.includes('Missing Text') || x['Text'] < textThreshold) &&
-                    (!completionFilters.includes('Delegation (1)') || x['Delegation'] >= 1) &&
+                    (!completionFilters.includes(`Delegation (${delThreshold})`) || x['Delegation'] >= delThreshold) &&
                     (!completionFilters.includes('No Delegation') || x['Delegation'] === 0) &&
                     (!completionFilters.includes('All Sim (4)') || x['Sim Count'] >= 4) &&
                     (!completionFilters.includes('Any Sim') || x['Sim Count'] >= 1) &&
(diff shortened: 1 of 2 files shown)


nextcen-dgemoets (approved):
My goodness.  There's a lot here and I didn't find any issues with the work performed.  I did all of the testing steps except for Step 6-- no extra credit for me.  I also really like the improvement to the Phase 1 progress table.

The only things I noted are not really to do with work performed in the ticket:
- The list of All Alignments (Highest to Lowest) lacks the decimal point in the alignment target id.  However, this is how the field names are laid out in the `userScenarioResults` table:  no decimal point.  Is `.` not a valid character in field names, perhaps?  I suppose this ticket could manually add the decimal point back in, but it's probably not worth it.
- We have not yet cleaned up the ADM names in the database (e.g., removing, at a minimum, the UUID in `ALIGN-ADM-Ph2-ComparativeRegression-Mistral-7B-Instruct-v0.3__d43bfed0-1efa-49d2-a41d-60030850d330`).  Is it safe to say that if this data is cleaned up in the DB, that this table will automatically reflect it?

dereknop:
> * The list of All Alignments (Highest to Lowest) lacks the decimal point in the alignment target id.  However, this is how the field names are laid out in the `userScenarioResults` table:  no decimal point.  Is `.` not a valid character in field names, perhaps?  I suppose this ticket could manually add the decimal point back in, but it's probably not worth it.

Correct, Mongo does not like '.' in keys. I already had a utility function elsewhere to add the decimal, so I just added that in now.

joshuale-caci:
Testing went well and it looks good. I was wondering if the excel file needs to also reflect the green cell color for the completed surveys. 
image

joshuale-caci (approved):
Just checked it again and they are now highlighted green in the excel! Nice work!

PR #317, opened 2025-07-24
New delegation materials

dereknop:
Run the [ingest pr](https://github.com/NextCenturyCorporation/itm-ingest/pull/150) first.

http://localhost:3000/review-delegation

You should be able to see the new observation materials under `Phase 2 July 2025 Collaboration`. You can spot check that the decisions displayed for an ADM fall in line with that document from the `admTargetRuns` collection. 

http://localhost:3000/results

The ADM results for July should now be showing up correctly if you go through and select a scenario/ADM/target combination (Only subset scenarios will be populated). You should also notice that the July ADM names have been trimmed.

### DIFF
--- a/dashboard-ui/src/components/Results/resultsTable.jsx
+++ b/dashboard-ui/src/components/Results/resultsTable.jsx
@@ -71,7 +71,7 @@ class ResultsTable extends React.Component {
         this.state = {
             adm: "",
             scenario: "",
-            evalNumber: 8,
+            evalNumber: 9,
             ADMQueryString: "history.parameters.adm_name",
             showScrollButton: false,
             alignmentTarget: null
@@ -173,7 +173,7 @@ class ResultsTable extends React.Component {
         if (evalNumber >= 4) {
             if (evalNumber === 7) {
                 return <MultiKdmaRq23 />;
-            } else if (evalNumber === 8) {
+            } else if (evalNumber >= 8) {
                 return <PH2RQ2223 evalNum={evalNumber}/>;
             } else {
                 return <RQ2223 evalNum={evalNumber} />;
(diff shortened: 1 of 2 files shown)


NeilDaniel07 (approved):
Everything works as expected. After running the ingest script, my dashboard review delegation page shows a new dropdown for `Phase 2 July 2025 Collaboration` ADM runs. Clicking on any of the individual runs for a particular ADM Name/Scenario Name/Alignment target combination brings up the ADM choices. Using Mongo compass, I was able to verify that the ADM Choices displayed on the dashboard matched the choices recorded in the database collection. To be extra cautious, I clicked through the survey on a few individual results pages and no runtime errors occured as I completed these. 

The review page also shows matching ADM run results and the ADM names are trimmed in the selection panel and at the top of the page once a full set of options is selected. The `Adm Name` field in the `Start Scenario` block still shows the full, untrimmed ADM name. I'm assuming this is intentional, but still wanted to document this since it seems that trimmed ADM names are now used everywhere else.

nextcen-dgemoets:
Hey, looks great.  A few things:
- Do we need to change our Survey Version to Version 7?
- I'm getting "Reviewing undefined - \<insert ADM name\> - undefined" at the top of every delegation survey.  What is undefined?
- I love the clean ADM names.  Any chance we can do that for June also?
  - That way we can also get rid of the extra Baseline Aligned ADMs in `/results` (not to mention the "Random" ones).  Do we know which one is correct?

brianpippin:
> Hey, looks great. A few things:
> 
> * Do we need to change our Survey Version to Version 7?
> * I'm getting "Reviewing undefined - <insert ADM name> - undefined" at the top of every delegation survey.  What is undefined?
> * I love the clean ADM names.  Any chance we can do that for June also?
>   
>   * That way we can also get rid of the extra Baseline Aligned ADMs in `/results` (not to mention the "Random" ones).  Do we know which one is correct?

We only need to change to the new survey version when the survey itself is ready.  Derek just got the review ready in the PR but doesn't have the ADM loading all set up for the new July Collab, so not point in setting the survey version yet.

I do think we should make a new different ticket, perhaps assigned to @NeilDaniel07 to clean up the ADM names and remove the not needed ones.

I believe the undefined messages is because we aren't loading anything yet, and this is review, but I will Derek chime in.

brianpippin (approved):
LGTM - Once any other comments are addressed as needed.

PR #318, opened 2025-07-25
Dev to Main

brianpippin:


dereknop:
Things to check:
http://localhost:3000/review-delegation

nextcen-dgemoets:
There are certain scenario IDs in [ADM results](http://localhost:3000/results) that don't have any data (i.e., `Performer/ADM` is blank).  I believe this is because the holdout probes (full set) haven't been run yet, but those scenario IDs (e.g., `July2025-AF-eval`) are already in the list.

I guess this is minor.  They will be populated eventually.

dereknop:
> I guess this is minor. They will be populated eventually.

Yeah, those full scenarios were uploaded already to the `scenarios` collection, which is why they are under the drop-down there. On my next dashboard pr I can tack something on for the options to only appear there if there is at least one performer/adm and target combination that could be listed

nextcen-dgemoets:
I'm not seeing Merit Focus Set 1 Baseline Target 0.8, which was one of the missing ones yesterday.  Anybody have the same thing?

dereknop (approved):

NeilDaniel07 (approved):

PR #322, opened 2025-07-30
Handle Duplicate Emails

kaitlyn-sharo:
On main, try to start a text scenario through the login page using "123@123.com". You will be brought to an antiquated page that asks for a PID, mil/civ, and VR experience. Notice the PID in the url is not from our current eval. This was the intended functionality, but we need an upgrade:
- Within an evaluation, we don't want duplicate emails
- If a duplicate email is entered within an evaluation, the participant should be given the same PID in case they didn't finish the first time, or in the case of a race condition
- Between evaluations, we want to allow duplicate emails in the case of duplicate participants, with a new PID assigned to the same email for different evaluations (i.e. 123@123.com might have pid 201 for eval 4, but have pid 701 for eval 9. This is allowed. But it can never have 701 and 702 within eval 9)
- On main, we require the hashedEmail field to be unique across all participants (unless it's null). In this branch, we require the hashedEmail field to be unique across all participants within an evaluation, not within the entire collection

To test: run https://github.com/NextCenturyCorporation/itm-ingest/pull/152 first. Then test with different and duplicate emails to make sure no old functionality broke (i.e. new pids are still assigned to different emails) and the new functionality works (123@123.com will now give you a new pid because this is a new evaluation). 

In short, it should now be impossible to return to a previous text survey version, even if using the same email.

Additionally, timestamps are added to the creation of entries in the participant log now, as requested by Derek for easy cleanup of testing.

### DIFF
--- a/dashboard-ui/src/components/App/index.jsx
+++ b/dashboard-ui/src/components/App/index.jsx
@@ -10,7 +10,7 @@ import { isDefined } from '../AggregateResults/DataFunctions';
 // Components
 import ResultsPage from '../Results/results';
 import HomePage from '../Home/home';
-import { SurveyPageWrapper } from '../Survey/survey';
+import { SURVEY_VERSION_DATA, SurveyPageWrapper } from '../Survey/survey';
 import { TextBasedScenariosPageWrapper } from '../TextBasedScenarios/TextBasedScenariosPage';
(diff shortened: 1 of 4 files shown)


dereknop:
Yeah, it's fine. I'll need to update the ingest script that the Jenkins job points at anyways to now include the plog so I can account for the format there.

PR #323, opened 2025-07-30
Participant Progress Table Refetch Data fix

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?jql=assignee%20IN%20%2860ba425da547eb00686ee0ce%2C%20empty%29&selectedIssue=ITM-1070)

Currently, on http://localhost:3000/participant-progress-table, the `Refresh Data` doesn't correctly update the table. You can test this by switching to dev and having one tab open on the progress table, then starting the text-based scenarios in a separate tab. Despite a new PID being generated, if you hit `Refresh Data` it won't populate at the bottom of the table. 

If you follow those same steps on this branch, it should populate a new row with the new PID. You should also be able to see the table update when they complete the text-based scenarios, as well as the survey, without needing to refresh the whole page.

### DIFF
--- a/dashboard-ui/src/components/Account/participantProgress.jsx
+++ b/dashboard-ui/src/components/Account/participantProgress.jsx
@@ -1,4 +1,4 @@
-import React, {useCallback} from "react";
+import React, { useCallback } from "react";
 import '../SurveyResults/resultsTable.css';
 import '../../css/admInfo.css';
 import { Autocomplete, TextField, Modal } from "@mui/material";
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
LGTM.  I tried it on `main` and it didn't work unless I refreshed the browser page.
On the branch, I could ~~stalk~~ track my pid as I completed the text and then delegation surveys, which updated after every comparison.

I guess my only question is...if a browser refresh works, why do we need this feature/button/code?

kaitlyn-sharo:
> I guess my only question is...if a browser refresh works, why do we need this feature/button/code?

I added this a long time ago when I had no other tasks, was bored, and wanted to make a cool feature for Jennifer to use. The benefit is that your filters (and hidden/shown column choices) refresh on a browser refresh, but not on this button press.

dashboard-ui/src/components/Account/participantProgress.jsx
```diff
@@ -101,9 +101,10 @@ export function ParticipantProgressTable({ canViewProlific = false }) {
     const closePopup = () => {
         setPopupInfo({ open: false, pid: null, scenarioId: null });
     };
-    const sortData = React.useCallback(() => {
+    const sortData = React.useCallback((data) => {
+        console.log('hit sortData')
```
kaitlyn-sharo, line 105: probably want to remove this before merging?

kaitlyn-sharo (approved):

PR #325, opened 2025-07-31
RQ8 variable updates; added '-' for empty cells

kaitlyn-sharo:
RQ8 had slightly incorrect code for showing '-' for empty cells (while still not affecting the download, as that messes with analysis), so I fixed that

I also removed Tagging_performance from the variable list, as it was overlapping with Tag_ACC. Jennifer said to keep Tag_ACC over the other one and confirmed that they were duplicate variables.


nextcen-dgemoets:
I'm not seeing any `-` for empty cells on prod RQ8.    What am I missing?  Is it some other condition that isn't currently appearing in the table?

rq8

nextcen-dgemoets:
> I'm not seeing any `-` for empty cells on prod RQ8. What am I missing? Is it some other condition that isn't currently appearing in the table?

Oh, maybe it only happens after running `itm-1035-rq8-updates`.

nextcen-dgemoets (commented):
Looks good in my dashboard.  However the time format question is resolved (see ingest ticket), then maybe the time-related fields in the definitions table should be updated/clarified.

Also, it seems like triage performance is a little more than "% correct treatments applied" (which is what's written in the definitions).  Perhaps we should elaborate there...something like what you have in the ingest PR: 
- `correct tools applied / (misses + total tools applied)`

joshuale-caci (approved):
RQ8 table shows - for empty cells and the all the times look rounded up to 2 decimal places.

PR #326, opened 2025-08-01
ITM 1071

dereknop:
Update to http://localhost:3000/research-results/rq2 to support the July ADM runs. 

You should be able to select July from the drop-down (leaving June as the default for now). Test that you can view all the July ADM runs and spot-check a few of the alignment scores using the `admTargetRuns` collection. Please also check that if you use a filter on July, and then switch back to June, the filters reset.

### DIFF
--- a/dashboard-ui/src/components/Research/RQ2.jsx
+++ b/dashboard-ui/src/components/Research/RQ2.jsx
@@ -16,7 +16,8 @@ const ALLOWED_EVAL_OPTIONS = [
     { value: 4, label: 'Dry Run Evaluation' },
     { value: 5, label: 'Phase 1 Evaluation' },
     { value: 7, label: 'Phase II Experiment 1' },
-    { value: 8, label: 'Phase II June 2025 Collaboration'}
+    { value: 8, label: 'Phase II June 2025 Collaboration'},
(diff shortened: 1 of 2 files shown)


NeilDaniel07 (commented):
Looks good to me. All July ADM runs are present in the table when I selected July in the dropdown. When I applied filters and then switched back to the June tab, filters reset as intended. I was also able to verify that for a few selected rows in the table, both the aligned ADM and baseline ADM alignment score matched the document in the `admTargetRuns` corresponding to that attribute/alignment target/set combination. 

However, when I was scrolling through the table rows, it seemed as though sorting based on `Target` wasn't always working correctly. I attached a couple of screenshots showing examples below. I checked the June tab and it doesn't seem to have this problem (on both this branch and prod). The rows seem to be sorted by `Trial_ID`, but I'm not sure what this column represents (since I don't see it as a parameter in documents in the `admTargetRuns` collection). If the intention is to perform primary sorting based on `Trial_ID` and `Target` is used for secondary sorting, then I guess this isn't an issue. Just wanted to bring it to your attention as a precaution. 

Screenshot 2025-08-01 122622
Screenshot 2025-08-01 122639

NeilDaniel07 (approved):
The table rows are now ordered by `Target` value.

PR #327, opened 2025-08-01
added rq134 toggle for evals 8&9

kaitlyn-sharo:
You will have to rebuild to test due to a necessary server schema change that helps populate the adm alignment column

- Added July Eval to RQ1, 3, and Exploratory Analysis pages (Derek is working on RQ2)
- Added auto-choose latest eval on these three pages (before, it was hard-coded)
- Combined Eval 8/9 data on RQ 8 (so if you're on the july page or june page, you should see the same exact thing)
- Added toggle on June RQ 1/3/4 table to add July data
- Added toggle on July RQ 1/3/4 table to add June data

I'm not 100% sure what data should look like here, so for those of you who know, please pay extra attention to that!


kaitlyn-sharo:
> LGTM. Some good refactoring as well. There's no July RQ8 data yet, right? Because we haven't run the latest probe matcher on prod yet? I'm just noting that June and July are giving the same 15 June results that are on prod, with no July yet. But I think that's expected as I write this.

No July RQ8 data yet, you are correct

PR #328, opened 2025-08-04
Adm Probe Responses page

dereknop:
[Ticket](https://nextcentury.atlassian.net/browse/ITM-1080)

July data should now be populating. This also fixes an issue with the June/July `KDMA` column not populating. Please spot-check a few of the KDMAs using `results.kdmas` in the `admTargetRuns` collection. Also, please check that the previous evals are still working. 

Note: We only have the subset runs for July, so scenarios like `July2025-AF-MF-eval` won't be populated yet.

### DIFF
--- a/dashboard-ui/src/components/AdmCharts/admProbeResponses.jsx
+++ b/dashboard-ui/src/components/AdmCharts/admProbeResponses.jsx
@@ -429,6 +429,14 @@ export const ADMProbeResponses = (props) => {
                         </>
                     )}
                 </div>
+                {!currentScenario && currentEval && (
+                    <div className="test-overview-area" style={{ display: 'flex', alignItems: 'center', justifyContent: 'center' }}>
+                        <div style={{ textAlign: 'center', color: '#666' }}>
+                            <h3>Please select a scenario to view probe responses</h3>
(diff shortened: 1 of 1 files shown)


NeilDaniel07 (approved):
Looks good to me. I now see the July data populating and the issue of `KDMA` values not appearing on prod is resolved in this branch. This didn't seem to break the table or Excel downloads for previous evaluations. Spot checking `KDMA` values in the `admTargetRuns` collection also verified that the correct information was populating in the `KDMA` column for these ADM runs. Excel downloads for this new July data are also working as expected (both for downloads corresponding to just one particular ADM as well as all ADM runs for a particular Evaluation/Scenario combination).

kaitlyn-sharo (approved):
This works, however:
1. On most pages, if no data is showing up because not all required choices have been made, we say "please select a scenario/target/whatever". It may be nice to add this here
2. Also, if no data exists for a scenario, I think it would be nice to either not give that scenario as an option or to say "No data available for this scenario", just so nobody thinks it's an issue with the dashboard

dereknop:
@nextcen-dgemoets @kaitlyn-sharo 

You'll need to rebuild the GraphQL container, but only scenarios with results will populate now. Also added text if no scenario is selected.

PR #329, opened 2025-08-06
Itm 1048 ph2 survey results

kaitlyn-sharo:
Updated http://localhost:3000/survey-results for June/July (TABLE ONLY)

Requested:
- Get rid of TA1 and TA2 columns (only one of each now)
- Update attributes
- Change target column to only target value or (high/high etc) for multi-KDMA ADMs
- Change “Scenario” to “set”
- Add DM4 for comparison in multi-KDMA block
- Update demographics for Phase II flow and levels

What to look for:
- Added June/July option to survey results 
- Added an extra column for attribute to each block
- Updated target columns as requested
- Changed "Scenario" to "set"
- Added DM4 for comparison in all blocks, since multi-kdma can be in any block and order matters
- Added 3 Forced Choice (FC) question
- Added alignment column for each forced choice column, mainly for the multi-kdma FCs
- Fixed role filter and military filter for June/July
- Updated demographics for Phase II
- Updated variable definitions
- Make sure other tables still work as expected


nextcen-dgemoets:
It looks like maybe we need a ph2 version of the variable definitions spreadsheet?  We had to do that on another page or two.

jaudick:
This may be unrelated to this ticket but why is "doctor" the only lower case title in the Phase 2 role dropdown? 

image

jaudick (approved):
I compared this to development and I see the new Phase 2 table and all of the new Phase 2 additions you described in your pr description

kaitlyn-sharo:
> This may be unrelated to this ticket but why is "doctor" the only lower case title in the Phase 2 role dropdown?
> 

My guess is it's something a participant wrote in the "other" box

nextcen-dgemoets:
> > This may be unrelated to this ticket but why is "doctor" the only lower case title in the Phase 2 role dropdown?
> 
> My guess is it's something a participant wrote in the "other" box

Yep, I've seen that too and come to the same conclusion.

nextcen-dgemoets (commented):
Everything LGTM, just the one question about PID 7101 and 7102.

kaitlyn-sharo:
> Everything LGTM, just the one question about PID 7101 and 7102.

Looking at the progress table, those are test PIDs, which most tables (probably/hopefully all of the tables) exclude as they should not be included in analysis

PR #332, opened 2025-08-08
added june/july toggles to del block table

kaitlyn-sharo:
http://localhost:3000/research-results/exploratory-analysis 

On the delegation by block table for June, added an "include july" toggle.
On the delegation by block table for July, added an "include june" toggle.

Test that these work and that the toggles on the delegation by block for phase 1 still works, as well as switching between evals using the dropdown.


dereknop (approved):
Worked as you described, nice.

NeilDaniel07 (approved):
Looks good to me. These toggles function as intended by adding and removing additional data from the table on click. No functionality is broken in the equivalent toggles in the `Phase 1 Evaluation` tab of the dropdown. Switching tabs on the dropdown still ensures that the correct data is displayed in the table. Excel downloads also correctly contain the data in the table, based on whether the toggle is clicked or not.

nextcen-dgemoets (approved):
Looks good, both June->July and July->June.  Confirmed that junk test data is in both this table and survey results.

PR #337, opened 2025-08-21
Itm 1092 delete pid

kaitlyn-sharo:
The goal of this ticket is to allow ADMINS ONLY to delete participant data. Derek currently has a script that accomplishes this based on the date of testing, but this adds similar functionality to the UI.

The feature has been added to http://localhost:3000/participant-progress-table. 

Things to check:
- Non-admins should not see anything different. 
- Admins should now have an extra "Delete" column. 
- Delete functionality is only allowed for data that is at least 24 hours old (except sim data). The time is checked in the client code to show/hide the button, and also on the server side to prevent date manipulation. This check will make sure that we have the backed up data in case of an accidental deletion.
- You should need to type in the PID exactly as written in order to delete on the confirmation page
- Successful deletions should give a notification (bottom left corner), the table should refresh, and you should no longer see that data
- Text, Sim (raw and analyzed), Survey, and Participant Log data should all be deleted
- If anything goes wrong (play around with different ways to make this happen, like changing dates in mongo, or even changing the code to force an error on the client side or server side), an error message should appear in the bottom left corner that says the data was not deleted
- Nobody except for someone with a valid admin session should be able to access this new, very dangerous endpoint

### DIFF
--- a/dashboard-ui/src/components/Account/participantProgress.jsx
+++ b/dashboard-ui/src/components/Account/participantProgress.jsx
@@ -1,7 +1,7 @@
 import React, { useCallback } from "react";
 import '../SurveyResults/resultsTable.css';
 import '../../css/admInfo.css';
-import { Autocomplete, TextField, Modal, Box } from "@mui/material";
+import { Autocomplete, TextField, Modal, Box, Snackbar, Alert } from "@mui/material";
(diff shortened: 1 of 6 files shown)


kaitlyn-sharo:
> It would be nice if the `Download All Data` didn't include the `Delete` column

haha, great point! I will fix that

PR #339, opened 2025-08-21
ITM 910: Redesigned Results Page with New Features and Increased Available Information

NeilDaniel07:
**Original Ticket:** [Here](https://nextcentury.atlassian.net/browse/ITM-910?atlOrigin=eyJpIjoiOWIxMmFlNWEyMjU1NDgyMTgwMTYwZDc2NWFjZDU0MWYiLCJwIjoiaiJ9)

## Description
This PR provides a major redesign of the `/results` page of the dashboard, introducing a new minimalistic and sleek design, additional ADM run information displayed, features such as truncating long text and hiding empty fields in the parameters and response tables, and a new tree table format that makes nested JSON payloads with ADM information far easier to read and inspect. Previously, once an `Evaluation`, `Scenario`, `Performer/ADM`, and `Alignment Target` were all chosen using the selections on the left pane, one large table would appear. The header to this table contained only the `ADM Name` and `Alignment Score` values. The table itself was not user friendly, consisting of a list of blocks (one for each command in the ADM run), each with a representation of the raw JSON payload for the `Parameters` corresponding to that command. This showed all fields, even if they were empty. Furthermore, the user had to manually click on a caret next to each block to view the `Response` table corresponding to that particular command. On the whole, this layout was unresponsive to user preferences (ex: viewing empty fields or not), tedious to navigate through, and largely visually unappealing. Here are the key capabilities introduced in this PR that adds significantly to the function, structure, and design of the `/results` page:

- New run ADM information is displayed at the top of the page that was not previously present. This includes the `Scenario`, `Alignment Target`, as well as the KDMA values(s) for specific attributes in that ADM run (one for single kdma scenarios and two for multi kdma scenarios). 
- Dynamic summary grid containing the above information, each in individual summary cards. The grid responds to changes in the number of KDMA values for a particular ADM run through layout changes to ensure a clutter-free design in any case. 
- A `Hide Empty Fields` toggle is added. When selected, all empty values in both the `Parameters` and `Response` tables (including `null`, `-`, `n/a` , and `NaN` number values) are removed from user view in the tables. 
- A `Truncate Text` feature limits the text size for values in tables to one line, making these tables far more compact if the user goal is a quick scan of the `Parameters` or `Response` object, rather than an in-depth review of each value. If selected, value strings are limited to one line, with an ellipses at the end. Strings that do not take up the entire space in their row of the table remain unchanged.
- The `Commands` are now listed in a sleek scroll on the left side of the page. Clicking on any individual Command now renders both the `Parameters` and `Response` objects corresponding to that particular command.
- In this Commands list, the `Respond to TA1 Probe` commands now also contain information on which probe was responded to. The format of these probes matches the specifications requested by Jennifer in [ITM-1047](https://nextcentury.atlassian.net/browse/ITM-1047).
- The `Parameters` and `Response` objects are formatted as tree tables. Nested objects are rendered as an indented, two-column key/value tree with per-row caret toggles.
- Every complex table value (nested tables) has a `View` button that opens up a right-side drawer that shows the selected value rendered as a table of keys/values. This drawer also shows the raw JSON object. This drawer can be closed with the `Close` button at the top right or by pressing the `Esc` key. 
- New `Expand All` / `Collapse All` buttons to the right of the Response header can be used to expand/collapse the entire Response tree, opening or closing any arbitrary nested table structure. This works for both object and array-root payloads.

## Disclaimer
Currently, only the response object panel uses the expand all and collapse all controls. The reason for this is that the `Parameters` object is far less complex and does not appear to have any nested structures. Even though adding these controls to the `Parameters` object as well would offer greater future resilience if this fact changes, my opinion is that it clutters the UI unnecessarily given the present format.

## Sample Screenshot
image

## How To Test
1. Navigate to the Results page at http://localhost:3000/results and select an `Evaluation`, `Scenario`, `Performer/ADM`, and `Alignment Target` to load a test run. First, ensure that the layout matches what you see in the sample screenshot above. 
2. Ensure that the Scenario, ADM, and Alignment Target summary cards you see in the top grid match the options you have selected in the left pane. Then, click on the last Command in the scroll. This should be titled `TA1 Session Alignment`. Click on the `Expand All` button. You should be able to see the Alignment score and KDMA values (either one or two depending on whether your scenario is single-kdma or multi-kdma) in the `Response` table. Verify that the numbers you see in the table match the corresponding summary cards in the summary grid at the top of the page.
3. On prod (https://darpaitm.caci.com/results), select the same `Evaluation`, `Scenario`, `Performer/ADM`, and `Alignment Target` choices to load the same test run. 
4. Return to your local version. Click on several commands. Check to see that the selected command’s parameters and response load on the right. Then, compare the keys and values to what you see for the same command on prod. Ensure that there is no difference in the values present (some boolean values that appeared as empty cells on prod will appear as `false` in this branch, this is intentional). 
5. In a particular `Response` object, use the carets to open and close individual nested structures. Make sure that this works efficiently and without any issues.  
6. Choose a different `Response` object now. Find one that has a complex nested structure (I recommend ones corresponding to `Take Action` or `Change Scene` commands). Ensure that a `View` button appears for these complex children tables. Click on the view button. The right-side drawer should appear. 
7. Confirm the drawer opens with a properly formatted title (e.g., meta_info → Meta Info). The JSON content should be rendered in the same format as the parent tables at the top. The raw JSON should be present at the bottom. Ensure that the two match (if you have `Hide Empty Values` selected, then it is expected that null or empty values in the JSON will not show in the key/value pairs at the top. Try both the `Close` button at the top right and the `Esc` key on your keyboard. Both should close the drawer.
8. Toggle the `Hide Empty Fields` checkbox. Ensure that if it is selected, empty/null values do not appear anywhere across either table for a particular command. If unselected, check to see that these empty values where only the key is present are still rendered.
9. Next, manipulate the `Truncate Text` checkbox. If selected, long text should be truncated to a single line in the table. If not, all text should be visible, regardless of how many lines on the table the corresponding row now occupies.
10. When on a command with a nested response, click `Expand All` and verify the entire tree expands. All carets should now be facing down and there should be nothing left to open. Now, click `Collapse All` and confirm it resets so that all carets need to be expanded again.
12. Change the options in the left panel so that you have selected a long ADM name. Ensure that this doesn't break the format in any way. Neither the summary grid nor the tables should overrun the edge of the page. 

[ITM-1047]: https://nextcentury.atlassian.net/browse/ITM-1047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nextcen-dgemoets:
This is on `development` also, so not to do with this PR, but something's gone wrong with the ADM names.  Maybe something's happening when we run certain scripts more than once?  Does @dereknop  have any ideas?

adm_badname2

dereknop:
Noted, yea nothing to do with this pr. This definitely happened last week because script 089 ran multiple times due to it and the rq2 script taking ages with autosave turned on

nextcen-dgemoets:
> Noted, yea nothing to do with this pr. This definitely happened last week because script 089 ran multiple times due to it and the rq2 script taking ages with autosave turned on. On the radar and will fix.

[ITM-1101](https://nextcentury.atlassian.net/browse/ITM-1101)

[ITM-1101]: https://nextcentury.atlassian.net/browse/ITM-1101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dashboard-ui/src/components/Results/resultsTable.jsx
```diff
@@ -196,6 +258,403 @@ class ResultsTable extends React.Component {
         }
     }
 
+
+    isEmpty = (v) => this.deepIsEmpty(v);
+
+    deepIsEmpty = (v) => {
+        if (v === null || v === undefined) return true;
```
kaitlyn-sharo, line 265: for the future (**don't change what you've done here**), we do have an "isDefined" function that checks if a value is null or undefined. It makes it easier when you have ternary statements so you're not working with too many parentheses, &&s, or ||s. It does not make a difference in code like this, but can be useful for clarity and succinctness elsewhere, so I just wanted to point it out to you so you are aware.

dashboard-ui/src/components/Results/resultsTable.jsx
```diff
@@ -371,17 +833,149 @@ class ResultsTable extends React.Component {
...
+                                                                ))}
+                                                              </ul>
+                                                            </div>
+                                                          </div>
+                                                          {(() => {
+                                                            const hist = Array.isArray(testData?.history) ? testData.history : [];
+                                                            const idx = Math.min(this.state.selectedIndex, Math.max(hist.length - 1, 0));
+                                                            const sel = hist[idx] || {};
+                                                            const hasParams = sel?.parameters && Object.keys(sel.parameters).length > 0;
+                                                            const hasResponse = sel?.response !== undefined && sel?.response !== null && !(this.isEmpty(sel.response));
```
kaitlyn-sharo, line 934: Here's a line where `isDefined` could help with clarity and length, just as an example.

kaitlyn-sharo (approved):
I looked through the code and didn't notice any issues. However, I wonder if there are any functions like snakeCaseToNormalCase or isObject that you can move to a utils function, as we can use those in other places and they are not specific to resultsTable. I think there are a few nice functions you made here that could potentially be used elsewhere, and that would help with the file size and readability/organization.

I like what you did with the page, and I really like having the probe name in the command list like we discussed - it makes it super simple to navigate through.

dereknop (commented):
Love the new look. One thing that I think would be nice is the ability to go back to the parent object after clicking into a child object. For example, if I click `Meta Info` from inside `State` I can't go back up a level to view `State` without closing the popup.
Screenshot 2025-08-25 at 12 56 26 PM

NeilDaniel07:
> One thing that I think would be nice is the ability to go back to the parent object after clicking into a child object. For example, if I click `Meta Info` from inside `State` I can't go back up a level to view `State` without closing the popup. 

I added this functionality in the latest two commits. As the user explores further into the nested tree structure, the path of objects they have clicked through appears at the top of the `View` modal. Users can then click on the name of any of the ancestor objects to return any arbitrary amount of levels up in just one click. Let me know if you think this design is user-friendly or if you would prefer some alternative approach.

NeilDaniel07:
> However, I wonder if there are any functions like snakeCaseToNormalCase or isObject that you can move to a utils file, as we can use those in other places and they are not specific to resultsTable. I think there are a few nice functions you made here that could potentially be used elsewhere, and that would help with the file size and readability/organization.

That makes sense - I agree there are a lot of general purpose functions I've implemented in `resultsTable.jsx`. I've moved the majority of functions I added as well as some that were already present in this file to a utils file in the same subfolder (`dashboard-ui/src/components/Results/`). Please let me know if there are any other functions remaining in `resultsTable.jsx` that you would like me to move to a `utils.js` file.

dashboard-ui/src/components/Results/resultsTable.jsx
NeilDaniel07: Thanks for letting me know. I see that function definition in `DataFunctions.js` now and will be sure to import and use it in the future.

dashboard-ui/src/components/Results/resultsTable.jsx
NeilDaniel07: Yeah, I see how that would significantly reduce verbosity in the definition of `hasResponse`.

PR #341, opened 2025-08-26
Itm 1063 - Unit Test Update 

kaitlyn-sharo:
This joint ticket from Darren and me updates the unit tests for Phase 2. 

- Make sure you are on npm version 16 or later (I personally use 18)
- Set up the tests like normal (`npm run setup-tests`)
- Run the tests (`npm run test`)
- Go into `.env` and change `REACT_APP_TEST_SURVEY_VERSION` to 5 instead of 7. This will set the tests to run with Phase 1 versioning
- Run again (`npm run test`)

All tests should pass. 

You will notice that Phase 1 tests have 5 fewer tests than the Phase 2 tests - this is because the ADEPT phase 1 server is not currently up, so the server tests cannot currently run properly for ADEPT Phase 1. When the phase 1 server is reinstated, we can run these tests again


dereknop (approved):
Phase two tests:
Screenshot 2025-08-26 at 1 49 13 PM

After changing to `REACT_APP_TEST_SURVEY_VERSION` to 5:

Screenshot 2025-08-26 at 1 55 35 PM

No issues to report. For the ADEPT Qualtrix tests, it might be worth adding in a test to cover the CACI online only path added in this [ticket](https://nextcentury.atlassian.net/issues/?jql=textfields%20~%20%22consent%2A%22&selectedIssue=ITM-1039&wildcardFlag=true). I don't know if that is necessary or would fall under this ticket just thought I should mention it.

nextcen-dgemoets (approved):

kaitlyn-sharo:
> No issues to report. For the ADEPT Qualtrix tests, it might be worth adding in a test to cover the CACI online only path added in this [ticket](https://nextcentury.atlassian.net/issues/?jql=textfields%20~%20%22consent%2A%22&selectedIssue=ITM-1039&wildcardFlag=true). I don't know if that is necessary or would fall under this ticket just thought I should mention it.

Great, let's make a new ticket for that

PR #342, opened 2025-08-27
removed sim-3 from phase 2 progress table

kaitlyn-sharo:
Run https://github.com/NextCenturyCorporation/itm-ingest/pull/163 first.

Then check out the participant progress table. Nothing should change for phase 1, but phase 2 should only have Sim-1 and Sim-2 now (no Sim-3, and no eval_open_world in Sim-1 or Sim-2)


nextcen-dgemoets:
Should the `Sim Count` field turn green when it's 2?  I thought that Phase 2 did that at some point, but Phase 1 definitely still does.

kaitlyn-sharo:
Updated the sim count green cells for phase 2, also updated the download sim count green cells.

RQ8 had lost some data when I made the ingest change, so I updated ingest and updated the dashboard again to match it. There was data being put in eval_open_world that was not being put in the desert/urban entries, so I moved the data to the proper documents in mongo.

nextcen-dgemoets:
> > Should the `Sim Count` field turn green when it's 2? I thought that Phase 2 did that at some point, but Phase 1 definitely still does.
> 
> It does not on prod, but I could add that in pretty easily if we want to!

Right, I was saying that I thought it used to turn green somewhere along the way of phase 2 development, but it isn't now (on prod or otherwise).  But it turns green in phase 1, so I thought it should in phase 2 also.  :)
Thank you for fixing!

PR #344, opened 2025-08-28
added missing probe data to rq8

kaitlyn-sharo:
The dump_rq8 python script in the itm-extras repo gave Jennifer the probe responses for each of the phase 2 open world sims. This adds that data to the RQ8 table on the dashboard rather than having to run an extra script to get it.

You will need to run the 1104 ingest script before you see the probes populate in this table. Then, you should see a total of 12 probes - 6 for desert, 6 for urban.


jaudick (approved):
I see a total of 12 probes - 6 for desert, 6 for urban.

nextcen-dgemoets:
All of the data looks good.  My OCD is triggered by the inconsistent column names, e.g., `Urban Probe5_MF`, `Urban Probe3_AF`, and `Urban Probe5_AFMF`, but `Urban Personal_safety` and `Urban Search2`.  Could we change the latter two to `Urban Probe1_PS` and `Urban Probe_SS2`?  (I know there's only 1 PS probe.)  Or change them all to be in the form of `Urban MF Probe3`.

I'll send text for the other column definitions under separate cover.

NeilDaniel07 (approved):
Looks good to me. I saw 22 new columns in the RQ8 table when compared to prod. This included 6 columns for Desert probes and 6 columns for Urban probes. I downloaded the Excel file and had no issues viewing values in these columns. Finally, I checked the `Variable Definitions` table popup as well as Excel downloads, and the definitions for these new columns seem to be showing as expected.

kaitlyn-sharo:
Some semi-major code changes required for the name changes darren requested, so requesting re-review

PR #353, opened 2025-09-15
Dev to Main

brianpippin:
* headers

* proptrust, trust, 2 new multi kdmas

* variables

* remove unused delegation header for eval 10

* text result fix


dereknop:
The eval drop-down on http://localhost:3000/text-based-results should now be in descending order of eval numbers.

PR #355, opened 2025-09-19
Update endpoint name from dre to phase1

brianpippin:
Already updated the server running at 8087 on production to be running at tag 3.9.9 instead of running a DRE version.

Will be hard to test this potentially only changing redirect URL name for Derek and UK collect.


nextcen-dgemoets (approved):
Would you want me to hit this server with local phase 1 server and client just to see that I'm e.g. getting session alignment with no errors?  Hmm, I guess I can't because my server isn't running on prod...

Oh well...LGTM then!

PR #358, opened 2025-09-24
ITM 1103: CACI Prolific Data Collection Testing Suite

NeilDaniel07:
**Original Ticket:** [Here](https://nextcentury.atlassian.net/browse/ITM-1103?atlOrigin=eyJpIjoiYjExMDMxMmNlOGI1NDkzYjgzODhmMjZkYjJhNTMxNTAiLCJwIjoiaiJ9)

## Description
This PR introduces a new test suite covering the CACI Prolific entry path to the text-based scenario survey. Previously, the existing `AdeptQualtrix.test.jsx` file in `/dashboard-ui/src/__tests__/AdeptQualtrix.test.jsx` tested the core functionality of the survey. However, it did not test the Prolific consent entry gate, which displays a consent form that the user must agree to before accessing the survey questions. This method of entry also changes `PID` logic, so that it is now maintained via a pass-through from the URL as opposed to being determined via online generation. This branch introduces `CACIProlific.test.jsx`, which mirrors the structure and flow of the existing `AdeptQualtrix.test.jsx` testing suite, however it adds coverage for the new consent gate. New tests in the suite verify that this consent form appears when `caciProlific=true`. It simulates clicking `Agree` to the form conditions (which proceeds either directly to scenarios when `startSurvey=true` or to the Instructions page if this is not the case. It also tests `Disagree` by ensuring redirection to the prolific return URL (`https://app.prolific.com/submissions/complete?cc=C155IMPM`). The testing suite developed walks through the full scenario flow and is compatible with both Phase 1 and Phase 2 survey versions. Finally, the test ensures that the same `PROLIFIC_PID` value initially generated is held throughout all steps of the survey. This provides comprehensive coverage to the Prolific entry gate that was previously not present. 

## Disclaimer
When testing my work locally, I noticed that Prolific redirects unauthenticated users from `app.prolific.com` to `auth.prolific.com`. To get around this and ensure that the redirection to the correct prolific return URL occurs, I added logic that asserts the outbound request is to the exact expected return URL before allowing landing on any other pages in the Prolific domain. This keeps the test deterministic and independent of changes made by Prolific.

## How To Test

1. From the root of the repository, run the terminal command `npm run setup-tests`. 
2. To save time and only run the new testing suite (instead of all testing suites in the test folder), first `cd dashboard-ui`. Then, run `npm test -- src/__tests__/CACIProlific.test.jsx`.
3. Wait for the test run to complete. The entire test suite should pass, displaying `6 passes, 6 total`. 
4. To verify the test suite is compatible with a Phase 1 survey version, change the value of the `REACT_APP_TEST_SURVEY_VERSION` variable in `.env` to 5 instead of 7 (following @kaitlyn-sharo's logic in #341). 
5. Re-run the command `npm test -- src/__tests__/CACIProlific.test.jsx`. All tests in the suite should still pass.
6. _(Bonus)_ Inspect the contents of the `CACIProlific.test.jsx` file. Ensure that the tests are equally as comprehensive as in the ADEPT Qualtrix suite and that the consent gate is tested thoroughly.

### DIFF
--- a/dashboard-ui/.env
+++ b/dashboard-ui/.env
@@ -39,4 +39,4 @@ REACT_APP_SOARTECH_DRE_URL=https://darpaitm.caci.com/soartech_dre
 REACT_APP_EMAIL_SALT=j0Bpq6e1b0556pu935uBdu
 
 # changing this will automatically change the text version to a valid option
-REACT_APP_TEST_SURVEY_VERSION=5
+REACT_APP_TEST_SURVEY_VERSION=7
--- a/dashboard-ui/.env
+++ b/dashboard-ui/.env
@@ -39,4 +39,4 @@ REACT_APP_SOARTECH_DRE_URL=https://darpaitm.caci.com/soartech_dre
 REACT_APP_EMAIL_SALT=j0Bpq6e1b0556pu935uBdu
 
 # changing this will automatically change the text version to a valid option
-REACT_APP_TEST_SURVEY_VERSION=7
+REACT_APP_TEST_SURVEY_VERSION=5
--- a/dashboard-ui/src/__mocks__/testUtils.js
+++ b/dashboard-ui/src/__mocks__/testUtils.js
@@ -165,17 +165,26 @@ export async function startAdeptQualtrixSurvey(page) {
 }
 
 export async function startCaciProlificSurvey(page) {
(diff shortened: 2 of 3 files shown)


nextcen-dgemoets:
So far so good (steps 1-4).  Step 5 (`AdeptQualtrix.test.jsx`) gave a timeout on connecting to the SoarTech server, presumably due to the AWS outages we're having.  I'll try again later.

PR #360, opened 2025-10-01
PS-AF Text Results Fix

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?jql=assignee%20%3D%2060ba425da547eb00686ee0ce&selectedIssue=ITM-1141)

The PS-AF text results section (under September collab) was not downloading correctly. The probe names in the column headers were also not properly adhering to the format we had for the other scenarios. You should now see table headers such as `Probe-PS-AF-207-Resp`.

Check the `Download All PS-AF Results` spreadsheet as well as `Download Scenario Results` spreadsheet. Both should now populate with all of the participant responses. The prior download should also correctly denote which of the two PS-AF sets the participant saw (this can be checked using the `participantLog` collection ex. `{ ParticipantID: 202509110 }`)

### DIFF
--- a/dashboard-ui/src/components/App/setupUtils.js
+++ b/dashboard-ui/src/components/App/setupUtils.js
@@ -58,8 +58,7 @@ export function setupTextBasedConfig(data) {
     
     for (const config of data.getAllTextBasedConfigs) {
         let tempConfig = JSON.parse(JSON.stringify(config));
-        
-        if (hasTextBasedImages && tempConfig.eval !== 'mre-eval') {
+        if (hasTextBasedImages && tempConfig.eval !== 'mre-eval' && tempConfig.pages && Array.isArray(tempConfig.pages)) {
             for (const page of tempConfig.pages) {
                 for (const el of page.elements) {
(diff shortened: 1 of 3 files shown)


NeilDaniel07 (approved):
Looks good to me. The column headers all look fine, using the correct probe naming format (both in the table and in the downloaded `.xlsx` file). I see all participant responses in the downloaded files. I also didn't notice any issues with the set number in the `All PS-AF Results` download when spot-checking against the `participantLog` collection.

nextcen-dgemoets:
Maybe this is deliberate / okay, but when I download, e.g., `text_result_data_Sept2025-MF1-eval.xlsx`, values that are 0 in the table are just blanks in the spreadsheet.

nextcen-dgemoets:
> Maybe this is deliberate / okay, but when I download, e.g., `text_result_data_Sept2025-MF1-eval.xlsx`, values that are 0 in the table are just blanks in the spreadsheet.

@brianpippin   Is this okay, or should this be fixed?

PR #362, opened 2025-10-02
Itm 1122

dereknop:
[Ingest PR](https://github.com/NextCenturyCorporation/itm-ingest/pull/174) first

Here is a zip of the Soartech server to set up locally (Instructions in the `README.md`). This will be needed to get scores back from the Soartech text-based scenario. 
[ta1-server-mvp.zip](https://github.com/user-attachments/files/22661013/ta1-server-mvp.zip)

Here is the 3.9.9 tagged ADEPT server we need. Follow the instructions for running with Python and set the port flag as `--port 8087`
[adept_server-3.9.9.zip](https://github.com/user-attachments/files/22661774/adept_server-3.9.9.zip)

I believe the necessary scenario files should already be included in these zips. Let me know if you run into trouble setting these up.

Edit your `.env` file to point to your local server locations:
Screenshot 2025-10-03 at 12 44 21 PM

Rebuild your dashboard container.

Go to http://localhost:3000/admin. Adjust your configuration to these options (the PIDs don't are not relevant for now). Leave the styling as `Updated Styling`.

Screenshot 2025-10-02 at 9 50 21 AM

After these configurations are set, you can view http://localhost:3000/review-text-based and http://localhost:3000/review-delegation

You can start the experiment by entering a test email [here](http://localhost:3000/login):
Screenshot 2025-10-02 at 1 45 46 PM

You should be presented with four scenarios (in random order): MJ5, IO1, MJ1, and VOL2.

Upon completion, you should be able to view your results in the `userScenarioResults` selection (you can use the pid in the URL to query the collection ex. `{ participantID: "202510106" }`). You should see the `mostLeastAligned` data as well as their kdma scores.
Screenshot 2025-10-02 at 1 53 34 PM

Next up, delegation http://localhost:3000/survey! Enter your PID. I have left a console log to provide more debugging information about the pages loaded. 

Screenshot 2025-10-02 at 2 04 21 PM

By expanding this information, you can verify if the correct pages were loaded or not. Using the Mongo Compass screenshot above as an example, you can see that the most aligned target for this test run was `vol-synth-HighCluster-ph1` and the least aligned was `vol-synth-LowCluster-ph1`. We can check that these targets were properly loaded using the console log: 
Screenshot 2025-10-02 at 2 07 57 PM

As you can see, a baseline was loaded and the most and least aligned targets are what we would expect. 

Any case of the most aligned or least aligned target not being loaded should be a result of overlap between them and the baseline. This should also be visible in the console log.
Screenshot 2025-10-02 at 2 10 24 PM

In this example, the least aligned target was `ADEPT-DryRun-Moral judgement-08` but since that overlapped with the baseline, `ADEPT-DryRun-Moral judgement-07` was loaded since it was the next least aligned target to the participant that did not overlap with the baseline. 

Click through the survey, checking that you don't find any accidental US/American references instead of British/UK and that the images and information are loading as expected. Then check that your responses were saved in the `surveyResults` collection.

### DIFF
--- a/dashboard-ui/src/components/Survey/survey.jsx
+++ b/dashboard-ui/src/components/Survey/survey.jsx
@@ -528,23 +528,19 @@ class SurveyPage extends Component {
             return {};
         } else if (this.state.surveyVersion === "9.0") {
             const allPages = this.surveyConfigClone.pages;
-            console.log(allPages)
             const introPages = [...allPages.slice(0, 4)];
 
-            const matchedLog = this.props.participantLog.getParticipantLog.find(
-                    log => String(log['ParticipantID']) === this.state.pid
-            );
 
             const participantTextResults = this.props.textResults.filter(
                 (res) => String(res['participantID']) === this.state.pid
             );
+            console.log(participantTextResults)
 
             const allBlocks = [];
             const scenarioTypes = ['MJ', 'IO', 'VOL']
             for (const scenarioType of scenarioTypes) {
                 const block = createScenarioBlockUK(
                     scenarioType,
-                    matchedLog,
                     allPages,
                     participantTextResults
                 )
--- a/dashboard-ui/src/components/Survey/surveyUtils.js
+++ b/dashboard-ui/src/components/Survey/surveyUtils.js
@@ -798,11 +798,9 @@ export function getParallaxAdms(surveyVersion, scenario, ioTargets, mjTargets, q
             misalignedStatus = validAdms['misalignedStatus'];
             break;
         case 'vol-ph1-eval-3':
-            console.log(volTargets)
             cols1to3 = ['vol-human-5032922-SplitLowMulti-ph1'];
             set1 = ["vol-synth-LowCluster-ph1", "vol-human-8478698-SplitLowMulti-ph1"]
             validAdms = getValidADM(getAllVolTargets(surveyVersion), volTargets, cols1to3, set1, [], []);
-            console.log(validAdms)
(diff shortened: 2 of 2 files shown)


nextcen-dgemoets:
In MJ5 and possibly other places it refers to SPC Springer (for specialist).  That designation appears to be US-specific:
> No, the UK military does not use the abbreviation "SPC" for specialists. The rank of Specialist (SPC) is specific to the U.S. Army and does not exist in the British Army's rank structure. 

[See more](https://www.google.com/search?q=does+the+uk+military+use+the+abbreviation+spc+for+specialists).

dereknop:
> In MJ5 and possibly other places it refers to SPC Springer (for specialist). That designation appears to be US-specific:


Fixed on the ingest side

nextcen-dgemoets (approved):
Text and delegation review looks good.
I ran through the text scenarios, and saw the `mostLeastAligned` data and `kdma` scores in the database.

Because I am not able to stand up a local SoarTech server due to Zscaler, I was only able to test the ADEPT delegation survey (thank for the work-around, Derek).  I'll be sure to check it on prod when deployed.

It looked like we were asking about VR twice.  At the start of the survey we ask, `"How much experience did you have with VR Gaming before today?"` and ask about queasiness, etc.  Then after the survey we ask, `"How would you describe your experience with virtual reality technology"`.  The difference here is "VR gaming" vs. "VR technology?"

brianpippin (approved):

PR #425, opened 2026-02-19
ITM-1234: Sticky evaluation selector

Garycheng92:
The evaluation selector on pages RQ1, RQ2, RQ3, and Exploratory Analysis will now persist between those pages.

To test:

Navigate to a research page and select an evaluation from the drop down:
image

Navigate to a new research page and confirm that the evaluation selector is still the evaulation you selected from the previous page, in this case RQ1 and RQ2 both have the June evaluations so the selection persists:
image

Also test selections that are not present in other pages to see if the selector defaults to the most current evaluation.

I did explore adding this functionality to other parts of the dashboard but that can be added on in the future if requested. For now this will only affect the Data Analysis grouping

### DIFF
--- a/dashboard-ui/src/components/Research/ExploratoryAnalysis.jsx
+++ b/dashboard-ui/src/components/Research/ExploratoryAnalysis.jsx
@@ -11,16 +11,19 @@ import { HumanVariability } from './tables/humanVariability';
 import { BlockedTable } from './tables/BlockedTable';
 import { CalibrationData } from './tables/CalibrationData';
 import { PAGES, getEvalOptionsForPage } from './utils';
+import { useSelector, useDispatch } from 'react-redux'
+import { setSelectedResearchEval } from '../../store/slices/configSlice';
 
 export function ExploratoryAnalysis() {
     const evalOptions = getEvalOptionsForPage(PAGES.EXPLORATORY_ANALYSIS);
-    const evalRQ1Options = getEvalOptionsForPage(PAGES.RQ1); 
+    const evalRQ1Options = getEvalOptionsForPage(PAGES.RQ1);
+    const dispatch = useDispatch()
+    const storedEval = useSelector(state => state.configs.selectedResearchEval) 
 
-    const [selectedEval, setSelectedEval] = React.useState(evalOptions[0].value);
+    const selectedEval = storedEval ?? evalOptions[0].value
     function selectEvaluation(target) {
-        setSelectedEval(target.value);
+        dispatch(setSelectedResearchEval(target.value))
     }
-
     // Extract all evaluation values that have an RQ1 page to conditionally render RQ4 content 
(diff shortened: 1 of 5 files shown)


NeilDaniel07 (commented):
Functionally, everything seemed to work for me. The choice of evaluation that I selected in one of the `Data Analysis` pages consistently persisted when switching to another page when that same evaluation was an option. I also checked table data to ensure that it always reflected the evaluation in the drop down. Further, in cases where a certain evaluation is available for one page but not for another, switching to this page correctly results in the default evaluation results being displayed. 

There is one _edge case behavior I ran into_ that I wanted to point out. When a particular selection `a` exists on a page but not on another page that is switched to, then the default evaluation for that page `b` is correctly selected. However, when switching back to the original page, then it's selected evaluation is `a`, even if `b` is a valid evaluation on this original page. For example:

1. Select `Multi-KDMA experiment` on the RQ2 page
2. Switch to the RQ3 page and `Phase II July 2025 Collaboration` is selected. 
3. Switch back to the RQ2 page and `Multi-KDMA experiment` remains selected. 

I'm not sure if this is intended, defined behavior or not, so I wanted to make it known before explicitly approving.

nextcen-dgemoets (approved):
Although I think the solution has some drawbacks-- and this is the fault of the ticket writer, not the dashboard developer-- I think it's an improvement and worth merging as is.

I wrote up the proposed alternate solution discussed herein in [ITM-1257](https://nextcentury.atlassian.net/browse/ITM-1257).

brianpippin:
I think we should have design discussion before committing this ticket. I liked and designed the original behavior which was it should default to whatever the newest (or current evaluation) is.  Most people analyzing aren't going to be going backwards to look at past data, so I always liked that it defaulted to most recent.  We should run this by Jennifer as her and ADEPT are the main users of the dashboard for these tables I would say.

nextcen-dgemoets:
> I think I would much prefer you are turning on a whole eval and each pages handle [disclaiming that no data exists yet for that eval]. Then we could stick with the latest eval. That is my opinion though and we should do some design/workshop on this with Jennifer to find preferred.

For what it's worth, I like that solution, too-- even better!

If that's not already a ticket, I could re-purpose 1257 for that-- or change it to a "discussion only" ticket that references all solutions on the table.

Garycheng92:
> > and
> 
> I would rather this be solved with some different things that Kaitlyn coded with turning on every page individually. I would rather see every eval in every drop down, and if no data exists for that eval or doesn't exist yet, there was a message on the page that there isn't date for this evaluation yet. I think the original solution of turning on pages individually was a neat approach, but I think I would much prefer you are turning on a whole eval and each pages handle that. Then we could stick with the latest eval. That is my opinion though and we should do some design/workshop on this with Jennifer to find preferred.

Yea I agree with this approach, I'll hold off on merging this ticket until we have a consensus on what we want to do.

PR #452, opened 2026-04-07
Itm 1319: Make email required for create account

dereknop:
CACI did some pen testing and was able to create some accounts without email addresses. This caused the admin page to crash on production due to undefined references. This pr adds in some option chaining so the page doesn't crash (we can then delete these accounts from the production admin page). 

I also added in some server-side validation. We were only validating on the client side, which is presumably how they exploited the system.

Go to https://darpaitm.caci.com/admin and enter your password. The page will crash.

Go to http://localhost:3000/admin and enter your password. It will not.

### DIFF
--- a/dashboard-ui/src/components/Account/adminPage.jsx
+++ b/dashboard-ui/src/components/Account/adminPage.jsx
@@ -405,9 +405,10 @@ function ApprovalTable({ unapproved, updateUnapproved, caller }) {
                     </thead>
                     <tbody>
                         {unapproved.map((user) => {
+                            console.log(user)
                             return (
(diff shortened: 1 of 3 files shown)


nextcen-dgemoets (approved):
On the feature branch using the provided database, I was able to log in with my admin credentials and delete the spurious admin requests.

It's great that they coordinated their penetration testing with us in a collaborative manner so as to minimize impact on the team and its deadlines.  Maybe they can launch their simulated DOS attacks during data collection.

Garycheng92 (approved):
Admin page is working for me on this branch.

PR #454, opened 2026-04-13
Delegation Bug Fix

dereknop:
We accounted for cases where the Oracle ADMs were the same across the two subpopulation groups at the same target. In this case, the next most aligned target is loaded for the other subpopulation group to avoid showing duplicate ADMs. However, we ran into an edge case today where the next most aligned target also shared the same responses. This will dynamically check every time that a survey is loaded that there are different responses, it will continuously walk down the most least aligned until it finds the next most aligned unique ADM. 

Changes:
- Deleted set of targets that are the same across the two subpopulations
- added `haveSameResponses` function to check the Oracle loading every time a survey loads and walk down participant's  mostLeastAligned as needed.

### DIFF
--- a/dashboard-ui/src/components/Survey/surveyUtils.js
+++ b/dashboard-ui/src/components/Survey/surveyUtils.js
@@ -1934,6 +1934,16 @@ const genComparisonPagev10 = (aligned, baseline, misaligned) => {
     };
 };
 
+const haveSameResponses = (page1, page2) => {
+    const rows1 = page1?.elements?.[0]?.rows || [];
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
I loaded the provided DB backup that showed the phenomenon with pid `202604106`.  I change that pid's subgroup to B, deleted the survey, re-took the survey, and found that the `April2026-AF-observe` scenario walked down the alignments to find a unique one, since AF-7 subpop A is the same as AF-8 subpop B.

PR #458, opened 2026-04-21
ITM-1326 Create TCCC Table

Garycheng92:
Grab the latest copy of mongodb from s3
Rebuild your GraphQL container

TCCC Table
1.  Navigate to http://localhost:3000/research-results/tccc verify the data in the table populates with the prefixed columns.
2. Verify only results from the selected evaluation appear.
3. Download a file from all options in the dropdown and verify that it contains the data that is shown on the table

Admin Page
1. Check that the TA3 input box is showing and roles are able to be assigned

TA3 Role
1. Create two test users assign one a TA3 role and the other an experimenter/evaluator/adept role. Log in as the user with the TA3 role, verify you can see and interact with the TCCC page.
2. Log in as the other use and verify you cannot see the TCCC link and you are redirected from http://localhost:3000/research-results/tccc
3. Feel free to try out different combinations of user roles to see if the TCCC link is accessible


nextcen-dgemoets (approved):
LGTM, for the most part.  The data was all there, the download worked as described in the PR, and the TA3 permission did what it was supposed to.  I think I had to refresh my browser to see the TCCC link as the TA3 user, but that's fine.

I don't think any of this matters because the page is internal only, but I'll mention the following:
- I assume it's correct that there is so little data for September (4 rows) and October (3 rows)?
- In the tables, some cells are empty, others have a dash, others are N/A.  What's the difference?  Also, I seem to recall Jennifer doesn't like dashes in cells.
- There are no variable definitions, but I guess that's because this table has such a narrow audience (Jennifer, Ewart, maybe Ashu?) that definitions aren't necessary.
- Usually when we have filters on a page, we have a `Download` and a `Download All Data` button.  I would have expected `Download All Data` to download all the data, not just what I'd selected in the dropdown.  Again, I'm sure nobody cares...
  - [Edit: but if we're only going to have one button that downloads the data for the current filter, I'd call it "Download", not "Download All Data".]

dereknop (approved):
Nice, user permissions are working. Looks like you addressed Darren's comments as well. LGTM.

PR #461, opened 2026-04-28
ITM-1333: Survey Results Update and Table Virtualization

dereknop:

I have updated the `resultsTable` component to account for the new questions we implemented for the April experiment and updated the variable definitions accordingly. 

I took this opportunity to improve performance on these tables with hundreds of rows, as my local performance was becoming abysmal. I added the `react-virtual` package, so you will need to rebuild your UI container. This cuts down on how many nodes need to be rendered in the DOM. It will only render what the user can see, speeding up our load time and toggling between filters greatly. Be sure to check that when downloading all of the data is still there, even though it may not all be rendered in the UI.


joshuale-caci:
I ran into this 
image
image
Is this duplicated on your side?

dereknop:
> I ran into this 

Oops, I messed up when resolving merge conflicts. Thought I had already removed the duplicate, but I guess I hadn't pushed. It should be ok now.

joshuale-caci (approved):
Took a while to rebuild my dashboard but got it stop showing an import error and the variable definitions show up! Page also feels faster.

Garycheng92 (approved):
New questions are showing along with the updated variable definitions, I really like the performance update I think it was badly needed for this table.

nextcen-dgemoets (approved):
Welp, I couldn't find any issues.  LGTM.
The scrolling behavior is a little different, but there's nothing wrong with it, so it's a good thing if it makes the page faster.

PR #463, opened 2026-05-01
Dev to Main

brianpippin:
RQ8 Table for April
Delegation Table Updates


Garycheng92 (approved):

patmole99:
I'm getting this error:

`Module not found: Error: Can't resolve '@tanstack/react-virtual' in '/usr/src/app/src/components/SurveyResults'`

Seems like the module may have just not been added to `package.json`. I'll see if I can push a quick fix.

patmole99:
Never mind, this looks like a Docker dependency-volume issue rather than a missing package. My `dashboard-ui` container was using a stale `node_modules` volume, so it didn’t have the dependency installed even though it was already in `package.json`.

Kayla mentioned she was seeing this error yesterday too, so we may just need to make sure we do a clean rebuild / refresh the `node_modules` volume after pulling the latest changes. Our normal dev rebuild script wasn’t enough in this case.

patmole99 (approved):

nextcen-dgemoets:
> Never mind, this looks like a Docker dependency-volume issue rather than a missing package. My `dashboard-ui` container was using a stale `node_modules` volume, so it didn’t have the dependency installed even though it was already in `package.json`.

Yes, I had to delete my containers / volumes to get things to run properly.

PR #464, opened 2026-05-04
Itm 1334: April Participant-Level Data

dereknop:
Go to http://localhost:3000/admin

Screenshot 2026-05-04 at 12 28 24 PM

Go to http://localhost:3000/humanSimParticipant

You can spot check the data using the `userScenarioResults`, checking the `combinedKdmas` field for a participant to see that their attribute scores are properly populating. 

Bonus: I fixed a CSS rule that I broke when doing my delegation results update. The alternating row colors were broken for tables that were not virtualized, which should now be fixed. You can check this out by looking at RQ8 on dev (or prod) vs. this branch.  

EDIT: The participant's kdma scores using the other subpopulation group have now been added. See the corresponding [ingest pr](https://github.com/NextCenturyCorporation/itm-ingest/pull/220).  The new fields appear next to the pre-existing kdma fields:
Screenshot 2026-05-05 at 11 59 11 AM


nextcen-dgemoets (commented):
LGTM.  I spot-checked two different PIDs and compared the kdma values in `userScenarioResults` to the table, and it all checked out.  The downloaded data looked good, too.

My only question is that the PR adds the `PH2_April_Variables.xlsx` file and I see where that file is referred to in `definitionTable.jsx`, but when I downloaded the definitions from the dashboard, it downloaded `ph2_aprilDefinitions.xlsx`.  Was it supposed to download `PH2_April_Variables.xlsx`?  What's the difference and which is correct?

nextcen-dgemoets:
Oh-- I also compared the RQ8 table in this branch compared to prod, and I see the style fix.
I did notice that I don't have RQ4 data in my local Exploratory Analysis page, but it was on prod.  I'm sure that's a me thing (DB out of date, an ingest script not run, a toggle not toggled, something like that).

dereknop:
> What's the difference and which is correct?

They are the same file, there is just some pre-existing logic that renames it when it is downloaded. I'll change it to preserve the original filename when it is downloaded.

nextcen-dgemoets:
> > What's the difference and which is correct?
> 
> They are the same file, there is just some pre-existing logic that renames it when it is downloaded. I'll change it to preserve the original filename when it is downloaded.

As long as it's the right content, I'm not concerned with the filenames.

nextcen-dgemoets (approved):

nextcen-dgemoets:
> EDIT: The participant's kdma scores using the other subpopulation group have now been added. See the corresponding [ingest pr](https://github.com/NextCenturyCorporation/itm-ingest/pull/220).  The new fields appear next to the pre-existing kdma fields:
Screenshot 2026-05-05 at 11 59 11 AM
-----


The new fields appear in the table and the download, but it looks like the new fields need to be added to the variable definitions?

nextcen-dgemoets (approved):
LGTM.  The variable definitions have been updated.

Now I'll be able to sleep tonight.

patmole99 (approved):
The CSS looks fixed and I can see the KDMAs and 'OtherSub' metrics.

PR #468, opened 2026-05-08
ITM-1347 Add server side filtering to OW Part 1

Garycheng92:
This PR speeds up the load time for the Open World Part 1 table in RQ8

Rebuild Graphql

- Go to http://localhost:3000/research-results/exploratory-analysis and https://darpaitm.caci.com/research-results/exploratory-analysis
- Make sure you have April selected from the dropdown
- Scroll to the Open World Part 1 table. It should load noticeably faster compared to prod

Double check that all 64 rows are still showing and the download functionality is still working.


nextcen-dgemoets (approved):
Wow, much faster!  Makes a world of difference.

dereknop (approved):
Good stuff. Query load time is pretty much instant for me now.

PR #469, opened 2026-05-12
Itm 1335: Progress Table Updates

dereknop:
This pr does some general refactoring of the participant progress table, along with some improvements to existing features. 

To test, point your `.env` at your local instance of the ADEPT server.

You should now be able to filter on participants with complete delegation surveys. This should now work across different evaluations regardless of the number of blocks in that eval's delegation survey.
Screenshot 2026-05-12 at 4 04 11 PM

Two blocks of logic that were repeated in several places were extracted into helper functions.

I added the ability to use the 'repair alignment' feature even if the alignment scores and kdma scores are not detected to be incomplete. During the last online-only data collection, we had a participant who was missing the necessary data to load a delegation survey, but was not detected as missing any data since it wasn't evident on the top level of their documents in `userScenarioResults`. This way, we can always fix/re-run a participant's alignment without needing to run an ingest script or anything like that. As always, the repair alignment feature should copy the logic/scoring of the given evaluation. If you are pointed at your local ADEPT server, you can test this out:
Screenshot 2026-05-12 at 4 10 32 PM

I added some progress messages, so if the server is slow on production, the progress is at least visible.

### DIFF
--- a/dashboard-ui/src/components/Account/participantProgress.jsx
+++ b/dashboard-ui/src/components/Account/participantProgress.jsx
@@ -118,8 +118,7 @@ export function ParticipantProgressTable({ canViewProlific = false, isAdmin = fa
 
     const getCompletionOptions = () => {
         const textThreshold = selectedPhase === 'Phase 2' || selectedPhase === 'UK Phase 1' ? 4 : 5;
-        const delThreshold = selectedPhase === 'Phase 2' ? 5 : selectedPhase === 'UK Phase 1' ? 3 : 4;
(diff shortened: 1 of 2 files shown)


NeilDaniel07 (approved):
Everything looks great to me. I tested filtering the progress table rows to only show participants with fully complete delegation surveys and this worked consistently across different `Phase` and `Eval` selections. The new `Repair Alignment` feature also functioned just as the PR description outlines. I was able to run this tool on any participant with complete `Alignment Status`. When testing locally, the repair ran relatively quickly and the logged progress messages matched the participant scenarios listed at the top of the pop-up. I think this is a really helpful feature and love the way it's been designed here.

nextcen-dgemoets (approved):
Love it-- it works great!  The filters worked, and I love the progress indicator on the recalculation.

Was the change from `Participant_ID` to `Participant ID` a different bug you fixed?

PR #470, opened 2026-05-15
ITM-1257 Revisit Page Stickiness

Garycheng92:
Previously, eval dropdowns on each data analysis page (RQ1, RQ2, RQ3, ExploratoryAnalysis) were driven by admin toggles, a new eval had to be manually toggled on per page before it appeared in the dropdown. This PR replaces that behavior with a query-driven approach where all evals always appear in every dropdown. Individual tables now handle the case where no data exists for the selected eval by showing "This table is not available for the selected evaluation."

To Test: 

Verify all dropdowns (RQ1, RQ2, RQ3, ExploratoryAnalysis) show all evals
Select an eval with no data for a table — confirm "This table is not available for the selected evaluation" message appears
Select an eval with data, confirm table renders correctly
Switch between evals, confirm no stale data flash

I wasn't sure whether to include the `Blocked Table` and `Calibration Data` table in the checks so I left it for now.

### DIFF
--- a/dashboard-ui/src/components/Account/editEvals.jsx
+++ b/dashboard-ui/src/components/Account/editEvals.jsx
@@ -34,8 +34,7 @@ const DELETE_EVAL = gql`
 
 const pagePaths = [
     '/', '/text-based-results', '/humanSimParticipant', '/humanProbeData',
-    '/human-results', '/results', '/adm-results', '/adm-probe-responses',
-    '/rq1', '/rq2', '/rq3', '/exploratory-analysis'
+    '/human-results', '/results', '/adm-results', '/adm-probe-responses'
 ];
 
 export function EditEvals({ caller }) {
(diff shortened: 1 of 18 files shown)


nextcen-dgemoets (commented):
> Verify all dropdowns (RQ1, RQ2, RQ3, ExploratoryAnalysis) show all evals
> Select an eval with no data for a table — confirm "This table is not available for the selected evaluation" message appears
> Select an eval with data, confirm table renders correctly

All good.

> Switch between evals, confirm no stale data flash

So, my understanding is that the page still should be "sticky", that is, if you're currently on, say, RQ2 with "Multi-KDMA experiment" selected (there is data), then switch to RQ3, you should still have "Multi-KDMA experiment" selected (because it's sticky) but you'll see "This table is not available for the selected evaluation."  (which is a sentence so it should have a period at the end, BTW).

Also, this new approach obviates the need for the checkboxes in the admin page, no?  Other than creating Evaluations (name, number) the rest of the behavior should be query-driven, no?  Or maybe the scope of this ticket is only RQ1, RQ2, RQ3, Exploratory Analysis, in which case only those checkboxes should be removed.

nextcen-dgemoets (commented):
Looks really good!  Within the scope of the Data Analysis dropdowns, it works great.

So, does this mean we can remove those checkboxes from the Admin page?

[Edit:  And write the ticket(s) to add this behavior for the rest of the UI pages?]

nextcen-dgemoets (approved):
LGTM.  I linked this ticket to the new one (ITM-1355).

dereknop (approved):
👍

PR #474, opened 2026-05-28
Itm 1354: Text-Based Scenario Loading Logic June 2026

dereknop:
Pull the latest ADEPT server. If you haven't already, populate it with the new scenario files Darren shared via Teams. 

Point your `.env` file to your local ADEPT server. If it wasn't already pointed there, you'll need to rebuild your Docker container. 

Go to http://localhost:3000/admin to make sure we have the right eval selected:
Screenshot 2026-05-28 at 1 34 55 PM

Take the text-based scenario from the login page or from `Test Text Scenario`. Six scenarios, each with six probes, should be loaded. After completing the text-based scenarios, we can check the results using the `userScenarioResults` collection. 

Screenshot 2026-05-28 at 1 29 18 PM

You should notice that the two 'trinary' scenarios only have alignment scores to trinomial targets 
Screenshot 2026-05-28 at 1 39 09 PM

And the same for the four binomial scenarios:
Screenshot 2026-05-28 at 1 44 40 PM

The AF and SS (binomial) documents should also have additional fields for AF-SS only scoring. These will be used to determine the delegation loading for the one 2D block of the delegation materials. These are the expected additional fields:
Screenshot 2026-05-28 at 1 46 05 PM

There is no delegation survey yet, and updates have not been made to the participant progress table. Those are both pending. If you want to, you can create an eval 17 on the admin page and toggle on the `/text-based-results`. You can view your results on http://localhost:3000/text-based-results.

### DIFF
--- a/dashboard-ui/src/components/Account/editEvals.jsx
+++ b/dashboard-ui/src/components/Account/editEvals.jsx
@@ -34,8 +34,7 @@ const DELETE_EVAL = gql`
 
 const pagePaths = [
     '/', '/text-based-results', '/humanSimParticipant', '/humanProbeData',
-    '/human-results', '/results', '/adm-results', '/adm-probe-responses',
-    '/rq1', '/rq2', '/rq3', '/exploratory-analysis'
+    '/human-results', '/results', '/adm-results', '/adm-probe-responses'
 ];
 
 export function EditEvals({ caller }) {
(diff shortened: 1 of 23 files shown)


nextcen-dgemoets (approved):
Everything LGTM!

To wit:
- [x] Six scenarios, each with six probes, should be loaded.
- [x] You should notice that the two 'trinary' scenarios only have alignment scores to trinomial targets
- [x] And the same for the four binomial scenarios
- [x] The AF and SS (binomial) documents should also have additional fields for AF-SS only scoring
- [x] If you want to, you can create an eval 17 on the admin page and toggle on the `/text-based-results`. You can view your results on http://localhost:3000/text-based-results

Note that the last one didn't work via the "Test Text Scenario" route, as these are filtered out.  But I redid it via logging out then "Text Scenario Login", and the results appeared in `/text-based-results`.  The downloaded data looked good too.

ktabasco-prog (approved):
It works for me! I see all 6 probes in userScenarioResults, AF and SS docs have additional fields for scoring, and there were two trinary scenarios. I also didn't see my results right away with /text-based-results, but following Darren's comment I fixed that issue.

PR #475, opened 2026-06-01
ITM-1365 Reorganize Exploratory Analysis Page

Garycheng92:
This PR reorganizes the Exploratory Analysis Page and Adds a new role to the admin page.

Changes:

1. Exploratory Analysis renamed to Open World
2. Open World ADM data previously in the Exploratory Analysis page has been moved to its own page called Open World ADMs
3. RQ4 is removed from the newly named Open World page.
4. The new `External Sim Researcher` role is created and only has access to the Open World and TCCC pages.

To check:

- Review the data Analysis tab to check that the correct names are being shown
- Check the Open World page and check to see if RQ4 has been completely removed
- View the Open World ADMs page to check for the Open World table.
- Create a new user and grant it the `External Sim Researcher` Role on the admin page. Log in with that user and see if you are only able to see the Open World and TCCC tabs in the Data Analysis tab.

### DIFF
--- a/dashboard-ui/src/components/Account/adminPage.jsx
+++ b/dashboard-ui/src/components/Account/adminPage.jsx
@@ -79,9 +79,15 @@ const UPDATE_TA3_USER = gql`
     }
 `;
 
+const UPDATE_EXTERNAL_SIM_RESEARCHER = gql`
+    mutation updateExternalSimResearcher($caller: JSON!, $username: String!, $isExternalSimResearcher: Boolean!) {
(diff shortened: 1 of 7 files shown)


patmole99 (approved):
The page names and data in the data analysis section now all look as described and I was able to create and log in with an 'External Sim Researcher' account. When logged into this account, only the Open World and TCCC tabs were accessible.

joshuale-caci (approved):
Had to inject a db backup but all the open world stuff populates correctly and the external sim researcher role was working!

PR #485, opened 2026-06-09
ITM 1381 - Exemption vs Normal Loading Bug

dereknop:
You'll need the most recent backup from S3, or message me and I'll send it.

There was a bug with ADM loading that should have been marked as exempt but was marked as normal. The actual ADM loading itself is correct due to overlap, but we need to denote it as exempt. This can easily be tested by comparing production to this feature branch:

https://darpaitm.caci.com/participant-progress-table Prod:
Screenshot 2026-06-09 at 11 50 25 AM

http://localhost:3000/participant-progress-table Local:
Screenshot 2026-06-09 at 11 49 12 AM


Garycheng92 (approved):
LGTM. ADM loading is showing up with `Exemption`

patmole99 (approved):
After checking out this branch, the ADM Loading column now shows "Exemption" instead of "Normal".

joshuale-caci (approved):
Exemption shows up!

PR #488, opened 2026-06-10
ITM-1371 Data Analysis Reorg

Garycheng92:
Summary of Changes:
- Moved all non-RQ8 research content (RQ5, RQ6, RQ7, HumanVariability, BlockedTable, CalibrationData) from Open World into a new Exploratory Analysis page
- Open World now shows only RQ8, with the eval dropdown filtered to only include evals that have open world data (evals 4–6, 8–10, 12, 15–17)
- Exploratory Analysis shows a "No data available" message for evals without misc dataset content (evals 15, 16, 17 and any future evals)

Grab latest db backup from S3

Review http://localhost:3000/research-results/exploratory-analysis and http://localhost:3000/research-results/open-world and check that no RQ8 data is in exploratory analysis and no other datasets besides rq8 are in the Open World page. 

The dropdown for Openworld should only show evaluations that have RQ8 data.

Because the dropdown for Openworld only shows certain evals, the eval sticky selector will default to the current evaluation on the Open World page if the user views an evaluation on another Data Analysis that isn't included in Open World and switches to the Open World Page

June 2026 Evaluation is pending the RQ8 data which is why it shows up in the Open World Dropdown

### DIFF
--- a/dashboard-ui/src/components/Account/editEvals.jsx
+++ b/dashboard-ui/src/components/Account/editEvals.jsx
@@ -162,7 +162,7 @@ export function EditEvals({ caller }) {
                 "rq1": true, 
                 "rq2": true,
                 "rq3": true,
-                "exploratoryAnalysis": selectedPages.includes('/exploratory-analysis'),
+                "exploratoryAnalysis": true,
                 "admProbeResponses": selectedPages.includes('/adm-probe-responses'),
                 "admAlignment": selectedPages.includes('/adm-results'),
                 "admResults": selectedPages.includes('/results'),
--- a/dashboard-ui/src/components/App/Header.jsx
+++ b/dashboard-ui/src/components/App/Header.jsx
@@ -91,6 +91,11 @@ export function Header({ currentUser, logout }) {
                             Open World ADMs
                             </NavDropdown.Item>
                         )}
+                        {isUserElevated(currentUser) && (
+                            <NavDropdown.Item as={Link} className="dropdown-item" to="/research-results/exploratory-analysis">
+                            Exploratory Analysis
+                            </NavDropdown.Item>
+                        )}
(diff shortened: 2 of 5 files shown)


nextcen-dgemoets (commented):
Overall very solid.  Just a few questions/comments:
- When I select Feb/April/June 2026, it presents a little label that says `No data available`.  It's a little austere.  Could we re-use something like the `This table is not available for the selected evaluation` we use in other places?
- Could we verify the `Delegation Data by Block` section is in the correct evaluations?  I'm guessing it doesn't belong in any of these places, but it's available in the following places with 357 rows with PIDs from June 2025 through June 2026:
  - October 2025, 4K, UK, Probe Set Size Experiment, Multi-KDMA, Metrics, Sept Milestone, and MVP.
  - Maybe we could just remove those entries from the drop-down?
  - Of course this could be a separate ticket, or maybe it's easy to address here.
- Speaking of `Delegation Data by Block`, but not specific to this PR, is that the `/exploratory-analysis` page behaves strangely sometimes:  select `Phase II KDMA Experiment`, `Eval 12 UK Phase 1`, `October 2025` and others.  In the `Delegation Data by Block` section, when you click on the close box of the evals, it disappears and then quickly reappears and the count returns to 357.  This may be fixed by addressing the above.
- Not specific to this PR, but is it known that in `/research-results/rq2`, June 2026 leaves you with a "Loading..." text, but others (like April 2026) leave you with "This table is not available for the selected evaluation." ?

nextcen-dgemoets (approved):
So the `Delegation Data by Block` table was removed from everywhere?  Why did it exist in the first place if `/survey-results` serves the same purpose?

Anyway, approved...

dereknop (approved):

PR #490, opened 2026-06-12
Itm 1380/1384: June 2026 RQ2 Updates

dereknop:
Note: This isn't the 'real' RQ2 stuff since more will be added to this table. This is just displaying all of the ADMs that were run on the observation materials. 

http://localhost:3000/research-results/rq2

This table should now populate for June 2026. You can spot-check some of these results using the `admTargetRuns` collection. 
Screenshot 2026-06-12 at 2 15 34 PM

These changes also fixed a bug where Feb 2026 would not load on production (it may still take ~10 seconds for it to load since that eval pulls down such a huge amount of ADMs). 

The forever 'Loading...' message should be gone from http://localhost:3000/results before you select an. ADM/Target combo. 
Prod:
Screenshot 2026-06-12 at 2 14 36 PM

Local: 
Screenshot 2026-06-12 at 2 14 01 PM

### DIFF
--- a/dashboard-ui/src/components/Research/tables/ph2_rq22-rq23.jsx
+++ b/dashboard-ui/src/components/Research/tables/ph2_rq22-rq23.jsx
@@ -42,7 +42,6 @@ export function PH2RQ2223({ evalNum }) {
     const [targetFilters, setTargetFilters] = React.useState([]);
     const [setFilters, setSetFilters] = React.useState([]);
     const [setConstructionFilters, setSetConstructionFilters] = React.useState([]);
-    const [targetTypeFilters, setTargetTypeFilters] = React.useState([]);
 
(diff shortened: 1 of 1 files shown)


joshuale-caci (approved):
Table appears for both and populate for the June eval and the February fix!

Garycheng92 (approved):
LGTM!

nextcen-dgemoets (approved):
LGTM!
- [x] http://localhost:3000/research-results/rq2 :  This table should now populate for June 2026.
- [x] You can spot-check some of these results using the `admTargetRuns` collection.
- [x] These changes also fixed a bug where Feb 2026 would not load on production
- [x] The forever 'Loading...' message should be gone from http://localhost:3000/results before you select an. ADM/Target combo.

PR #493, opened 2026-06-29
Itm 1387: RQ1

dereknop:
Run [ingest pr](https://github.com/NextCenturyCorporation/itm-ingest/pull/236) first

Go to http://localhost:3000/research-results/rq1 make sure June 2026 is selected. Check that the table starts populating and that the variable definitions look good. 

The most important rows to check are `Alignment score (Delegator|Observed_ADM (target))` and `Alignment score (DelegatorTRI|Observed_ADM (target))`. The AF-SS block won't have `Alignment score (DelegatorTRI|Observed_ADM (target))` but all of the other ones should. You'll also notice that rows are being marked as exempt as needed after our score updates in part 2 of the ingest script. 
Screenshot 2026-07-01 at 12 20 20 PM


joshuale-caci (approved):

patmole99 (approved):
Had to restore my db and point at prod adept, but after that my values matched Derek's.

PR #498, opened 2026-07-07
Dev to Main

brianpippin:
RQ2 and ADM Probe Responses


dereknop (approved):

nextcen-dgemoets (approved):
The new stuff looks good.

The only thing that caught my eye is that `/adm-probe-responses` doesn't have evaluation scenarios (but this is the same as Feb2026), so I guess it's by design.

nextcen-dgemoets (approved):
The full eval scenarios (with alignment score of `Not Requested`) are gone, including from the download.

joshuale-caci (approved):
All "Not Requested" scenarios are gone and it looks good!

PR #501, opened 2026-07-10
Itm 1394: Text Scenarios Guard & Resume Partially Completed Run

dereknop:
You'll want to point your `.env` file to your local ADEPT server and rebuild your dashboard container. Rebuild your Docker container even if you are already pointed locally, because I had to make some GraphQL schema changes. 

Prevents any user from taking the text-based scenarios with the same email address if they have a COMPLETE run already recorded for the current evaluation. 

If you try to take the text-based scenarios with the same email and already have a completed run, you should see:
Screenshot 2026-07-10 at 2 21 43 PM

The point at which different text documents get uploaded varies by eval, depending on which scenarios share session IDs. For example, in our current eval, the trinary documents do not get uploaded until both are completed (since their probe responses share a session ID). So if a participant completed one and then exited the browser before completing the other, we won't have either of them. I point this out because the way I chose to handle this was to have them retake both in that case. 

To test this, you can take a complete run and then delete some of their documents (delete both of the trinary documents):
Screenshot 2026-07-10 at 2 30 53 PM

Then update their participant log entry to only have 4 text entries (simulating that they clicked off during the trinary scenarios):
Screenshot 2026-07-10 at 2 31 37 PM

Then, if I go back to try to take the text-based scenarios again, I'll be prompted with the consent modal and then presented with the two scenarios I am missing:
Screenshot 2026-07-10 at 2 34 47 PM

Clicking through this as normal should result in the documents being scored as usual with the ADEPT server (can verify with `userScenarioResults`)

### DIFF
--- a/dashboard-ui/src/components/Account/participantProgress.jsx
+++ b/dashboard-ui/src/components/Account/participantProgress.jsx
@@ -60,7 +60,7 @@ const TEXT_THRESHOLD_BY_EVAL = { 13: 12, 17: 6 };
 // Phase 2 evals that use a delegation threshold of 4 instead of 5
 const DEL_THRESHOLD_4_EVALS = new Set([10, 16]);
 
-const computeTextThreshold = (evalNumber, isPH2OrUK) =>
+export const computeTextThreshold = (evalNumber, isPH2OrUK) =>
     TEXT_THRESHOLD_BY_EVAL[evalNumber] ?? (isPH2OrUK ? 4 : 5);
 
 const computeDelThreshold = (evalNumber, isUK, isPH2OrUK) => {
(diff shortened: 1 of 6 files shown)


nextcen-dgemoets (approved):
Awesome; LGTM.

- [x] If you try to take the text-based scenarios with the same email and already have a completed run, you should see \<pic\>
- [x] [If] I go back to try to take the text-based scenarios again, I'll be prompted with the consent modal and then presented with the two scenarios I am missing.
- [x] Clicking through this as normal should result in the documents being scored as usual with the ADEPT server

After re-taking the trinary scenarios, there were 6 records in `userScenarioResults` and the `textEntryCount` in the `participantLog` went back to 6.

Garycheng92 (approved):

PR #502, opened 2026-07-10
ITM-1355 Adding Additional Eval Sticky Selectors

Garycheng92:
This PR removes the admin per-page eval toggle in favor of query-based availability. Every eval shows in every dropdown, each page shows its own `no data available` message if it has nothing for the selected eval. It additionally migrates the sticky eval selector [from #470](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/470) to the pages that didn't have it yet.

Previously every time when a new eval was added the pages to display the eval had to be toggled. Now the pages are query based, each page will read from its queries whether it has data to display for the selected eval, if it does not, it will display a `no data available` message.

The admin eval editor no longer has per-page switches, it only has eval name, eval number, and delete. Since moving to a query based approach, these toggles are redundant.

Also fixes /human-results for evals 12 and 16, which had real data but were stuck on "please select a scenario and participant" with an empty scenario menu.

To test:

Click around every page and check that all evals are available in the dropdown, also click around the pages to ensure the same eval is being displayed throughout the dashboard. Look in particular for `no data available` messages flashing before the data loads.

Each page should show a `no data available` message or the contents of the data.

Create a new eval, refresh your page and check that it shows up in every page dropdown. 

(The programQuestions page will be overhauled so that page is not a part of this update and the `TCCC` page has its own data pipeline so it will be unchanged)

### DIFF
--- a/dashboard-ui/src/components/Account/editEvals.jsx
+++ b/dashboard-ui/src/components/Account/editEvals.jsx
@@ -2,20 +2,12 @@ import gql from "graphql-tag";
 import React from "react";
 import { useMutation, useQuery } from "react-apollo";
 import { Button, Card, Modal, Form } from "react-bootstrap";
-import { IconButton, Switch } from '@material-ui/core';
+import { IconButton } from '@material-ui/core';
 import DeleteIcon from '@material-ui/icons/Delete';
-import store from "../../store/store";
-import { setEvals as setEvalsInStore } from "../../store/slices/configSlice";
 
 const GET_EVAL_DATA = gql`
     query GetAllEvalData {
         getAllEvalData
-    }
-`;
-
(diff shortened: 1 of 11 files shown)


nextcen-dgemoets (approved):
LATM.  (Looks amazing to me.)

The only things I'll note:
- If you open a dropdown in another window (via middle click or right click -> Open in New Tab), that new window is set to the latest eval, not your current window.  I actually think this is a feature:  it enables you easily to have separate tabs with different evals open for comparison.
- One thing I noticed is that in some windows, it will scroll the evaluation dropdown to your current selection; in others, it did not.  It would be preferable if it did this everywhere, but nothing to hold up approval of the PR.

Nice work!

dereknop (approved):

PR #503, opened 2026-07-14
Itm 1413: /survey-results update

dereknop:

Eval 17 (June 2026) should now be supported. 

Screenshot 2026-07-14 at 1 24 58 PM

AF-Tri/AF-Bi and PS-Tri/PS-Bi should now be properly distinguished in the attribute column. Please spot check some data using `surveyResults` and `demographicsData`

Note: You'll want to rebuild your Docker container since I made a schema change (demographics data endpoint now can take an array of evalNumbers or a single evalNumber as an argument).


nextcen-dgemoets (approved):
LGTM.  The table looks good, and my spot check matched what was in the database.

The variable definitions had some old stuff (e.g., "Exclusive to the September Evaluation"), but I'm not hung up over it.

joshuale-caci (approved):
Data matched the table and the new eval 17 is avaiable for the it too!

Garycheng92 (approved):
I see eval 17 and the data matched what was in the table

PR #504, opened 2026-07-14
ITM-1415: Dev Server GraphQL Build Fix

patmole99:
This is a quick fix that makes it so the GraphQL server gets properly rebuilt when using docker compose commands. This branch is already checked out and built on the dev server, so to test just go to:

https://darpaitm.caci.com/dev/login

And attempt to login. If your login in successful, then everything is working properly.


nextcen-dgemoets (approved):
Yep, I was able to log in (and log out and log back in).  LGTM.

Garycheng92 (approved):
I was able to log in

PR #505, opened 2026-07-16
Itm 1414: April fixes for sever error

dereknop:
[Ingest pr first](https://github.com/NextCenturyCorporation/itm-ingest/pull/239)

First, let's check the participant-level data http://localhost:3000/humanSimParticipant

Screenshot 2026-07-16 at 1 24 17 PM

Spot check these values for a participant using `userScenarioResults`. You can find the data under `combinedKdmas` and `otherSubKDMA`.

Next, check out RQ1 http://localhost:3000/research-results/rq1
Screenshot 2026-07-16 at 1 26 11 PM

Check that all three of the alignment score columns are populating. At the end of the table, you should see that `Delegation Preference (AlignedSS/Misaligned)` has been replaced with `Delegation Preference (AlignedOS/Misaligned)`. This is because AlignedSS/Misaligned doesn't exist anymore. All of the participants' subpopulations flipped, so the forced choices we have recorded are between AlignedOS/AlignedSS and AlignedOS/Misaligned.
Screenshot 2026-07-16 at 1 27 04 PM


joshuale-caci (approved):
All 3 columns appeared and data was populated. Cross checked with mongoDB.

ktabasco-prog (commented):
I spot-checked the participant-level data against `userScenarioResults` and verified that the expected values are present in the database. I also confirmed that the RQ1 alignment score columns are populating correctly and that the Delegation Preference column has been updated to `AlignedOS/Misaligned` as expected. Everything LGTM!

The only follow-up I have is related to my comment on the ingest PR (ITM-1414): I noticed a few participants (e.g. `202604125` and `202604129`) that logged "no survey found" and don't have documents in the `human-to-adm-comparison` collection. If that was intended behavior and those participants were supposed to be skipped for that collection when no survey is found, I'm good with approving this PR.

patmole99 (approved):

PR #507, opened 2026-07-21
Itm 1362: Package.json cleanup and Performance Optimizations

dereknop:
This PR removes unused/unnecessary dependencies from our `package.json` file to cut down on the size of our build. The application loading is no longer blocked by the config files for the survey and text-based scenarios to be completely loaded (these are the most expensive queries on page load). 

Lazy loading/importing of our non-critical components has also been implemented. This means that if a dependency is only used on a results page, it does not need to be included in the bundle.js. For example, a participant coming to the site to take a text-based survey shouldn't need to wait for `xlsx-js-style` to be loaded (a dependency used for exporting result tables). `index.jsx` has been updated so that pretty much all of the results pages and all admin-only pages are now using lazy imports. 

Dev:
Screenshot 2026-07-21 at 2 55 39 PM

This branch:
Screenshot 2026-07-21 at 2 53 41 PM

I'm seeing about half the load time locally for the initial bundle.js between dev and the feature branch, so I am hopeful this will be pretty impactful on production. All behavior should be the same, besides the stated changes above. You'll need to rebuild your Docker container.

Note: I messed up rebasing this branch and had to fix some stuff manually, that is why the commit history has some duplication.


nextcen-dgemoets (commented):
LGTM for the most part.  I'm mostly seeing some graphical changes/anomalies:
- The variable definitions pages are oddly translucent instead of opaque.
- The home page dropdown spans the entire page.
- The three-way toggle on the /text-based-results page is styled different, and I like the change.

Also, can I / we delete the Test participants on prod?
202606100-202606102, although it says they're from the June 2025 Collaboration eval.

nextcen-dgemoets (approved):
The graphical stuff I mentioned looks fixed.  LGTM.

Unrelated, I noticed that `/humanProbeData` 's table for Eval 12 looks broken maybe.  It just says "ADEPT: Phase1-MJ4-eval, SoarTech: Incomplete" but there's no data anywhere.  I know we didn't do SoarTech in the UK, but I think the ADEPT data should be displayed?

Garycheng92 (approved):
Loads much faster now!

PR #508, opened 2026-07-23
ITM-1417 Add OW Part 2 table

Garycheng92:
Run [ingest](https://github.com/NextCenturyCorporation/itm-ingest/pull/240)

Rebuild GraphQL container

Navigate to http://localhost:3000/research-results/open-world-adms

The Open World Part 2 Table should be populated:
Screenshot 2026-07-23 154305

Spot check the table with `admTargetRuns` collection as well as reviewing the variable definitions.


dereknop (approved):

nextcen-dgemoets (approved):
Okay, took me a while to grok the data, account for why some data looked "missing," but I *think* it all checks out.

A few minor nice-to-have nits that won't prevent me from approving:
- The variable definitions could explain what "N/A" means in the `Tag_Expectant` column;
- The OW Scenario definition should mention more than just April;
- The ADM Name "Levels" description is wrong / out-of-date;
- The `Desert Probe_AFMF3` is at the far right of the table instead of grouped with the other Desert Probe columns; and
- The ADM name UUIDs have not been stripped.

PR #512, opened 2026-08-03
ITM-1427:  Fix empty PIDs that populate June 2025

dereknop:
PIDs that were generated with no data attached to them were populating under `June 2025 Collaboration`. On dev, you'll see 82 entries under June 2025. On the feature branch, this should be down to 66:
Screenshot 2026-08-03 at 4 47 17 PM

The fallback logic for when there was no data associated with the PID was not taking into account that we have an `evalNum` field that gets attached to the participant when they are generated. These garbage rows still need to be deleted (I will do this on production), but they will no longer congregate under June 2025 Collab.


nextcen-dgemoets (approved):
LGTM.  On the feature branch, there were only 66 June2025 rows.  The download reflected the same.

Garycheng92 (approved):

PR #513, opened 2026-08-06
Itm 1425: Virtualize RQ1 and RQ2 tables

dereknop:
This adds virtualization to the RQ1 and RQ2 tables. I think this will help with load times on production a lot. It should also limit lag when scrolling through the table itself. Only the rows visible in the viewport (and a buffer outside of that) will be rendered. This is opposed to every single DOM element of the table needing to be rendered on page load.

RQ1: http://localhost:3000/research-results/rq1

RQ2: http://localhost:3000/research-results/rq2

Check the code, that the tables still populate as expected, downloads work, etc.

### DIFF
--- a/dashboard-ui/src/components/Research/tables/ph2_rq22-rq23.jsx
+++ b/dashboard-ui/src/components/Research/tables/ph2_rq22-rq23.jsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { useVirtualizer } from '@tanstack/react-virtual';
 import '../../../css/resultsTable.css';
 import { useQuery } from 'react-apollo'
 import gql from "graphql-tag";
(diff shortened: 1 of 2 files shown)


nextcen-dgemoets (approved):
Well, I don't really know what I'm looking at / looking for, but I'm willing to say that things are better, and I didn't see anything broken.

I used the Performance tab of Developer Tools, and the number for "Interaction to Next Paint" in particular were a lot better on the feature branch than the development branch.

I couldn't necessarily tell a difference just clicking around the page, but this is running locally-- it might be more obvious on production, let alone from a slow Internet environment.

patmole99 (approved):
The data looks good and loads and downloads efficiently.

PR #514, opened 2026-08-07
ITM-1412: React Error Boundary 

ktabasco-prog:
**Issue**

- Expands upon ITM-1364: Error handling component by implementing a global React Error Boundary that wraps the entire application. This ensures that unexpected JavaScript runtime errors that would otherwise crash the app are caught and handled gracefully.
- The existing Error Boundary component has also been updated to invoke the Error Handling Modal introduced in ITM-1364, providing users with consistent, descriptive error messaging regardless of where an unhandled application error occurs.

**Test Instructions**

1. Verify the application loads normally and all primary navigation and functionality continue to work as expected.
2. Navigate to several pages throughout the application and confirm normal functionality when no errors occur.
3. Introduce an obvious JavaScript runtime error within a rendered component (for example, throw new Error("Test Error Boundary") during render or reference an undefined property that causes a runtime exception). 

- For an easy error test, the app still has the  .includes() error in the Metrics Evaluation tab under RQ2 of Data Analysis dropdown. This error was intended to be fixed in a separate ticket (ITM-1395).

4. Verify the application does not crash or display a blank screen.
5. Confirm the Error Boundary catches the exception and displays the Error Handling Modal introduced in ITM-1364.
6. Repeat testing from different areas of the application to ensure the global Error Boundary consistently catches unhandled runtime errors regardless of where they occur.
7. Confirm that after removing the simulated JavaScript error, the affected page renders normally and the rest of the application continues to function as expected.

**Expected Result**

Unexpected JavaScript runtime errors anywhere in the application are caught by the global React Error Boundary. Instead of crashing the application, users are presented with the Error Handling Modal and a clear, descriptive error message, providing a consistent and recoverable error experience across the entire application.


nextcen-dgemoets:
Not sure if Kayla can or should be reached for this, but should it be reassigned and investigated?

PR #516, opened 2026-08-12
Itm 1428: New Open World ADM runs

dereknop:
[Ingest pr first](https://github.com/NextCenturyCorporation/itm-ingest/pull/244)

http://localhost:3000/research-results/open-world-adms

New runs should be visible. You can check what you see using the `admTargetRuns` collection. Check that you can toggle the inclusion of the old runs (regression scoring) on and off. When they are excluded, only the columns actually populated with data should appear in the table, so you shouldn't see intercept, medical, or attribute.


nextcen-dgemoets (commented):
Looks good.  The filter works, and the table and corresponding table are updated appropriately.

The new column headers should match in OW1 and OW2:  In OW2, `affiliation` should be `Affiliation Value` like in OW1; likewise for `merit`.

And, um, the Variable Definitions could stand to be updated:  add `Affiliation Value` and `Merit Value`, and add new targets (e.g., `ADEPT-June2025-affiliation-*` or `ADEPT-June2025-merit-*`) to the `Target` column.

nextcen-dgemoets (commented):
An improvement!  But it looks like `Definitions_RQ8_OW_Part1.xlsx` still needs a bunch of new variables (e.g., Affiliation Value, AF_attribute, AF_intercept, AF_medical, Merit Value).  Some of these I guess are from a previous ticket.

nextcen-dgemoets (approved):

Garycheng92 (approved):

